PR #6, opened 2019-10-29
Fail vulcan-zap only on internal errors

julianvilas:


cmd/vulcan-zap/main.go
```diff
@@ -70,7 +71,7 @@ func main() {
 		}
 		client, err := zap.NewClient(cfg)
 		if err != nil {
-			return err
+			return fmt.Errorf("error configuring the zap proxy client: %v", err)
```
gguillemas: I will buy you a beer if you can humor me and upper case ZAP.

cmd/vulcan-zap/main.go
```diff
@@ -87,21 +88,52 @@ func main() {
...
+		}
+
+		v, ok := resp["scan"]
+		if !ok {
+			// Scan has not been executed. Due to the ZAP proxy behaviour
+			// (the request to the ZAP API does not return the status codes)
+			// we can not be sure whether it was because a non existant target
+			// or because an error accessing the ZAP API. Therefore, we will
+			// terminate the check without errors.
+			logger.WithFields(logrus.Fields{"resp": resp}).Error("Scan not present in response body when calling Spider().Scan()")
```
gguillemas: Maybe we discussed this and I forgot, but I would just log with "Warn" or even "Info" level here as we are not returning an error.

cmd/vulcan-zap/main.go
```diff
@@ -114,51 +146,85 @@ func main() {
...
-			a := alert.(map[string]interface{})
+			a, ok := alert.(map[string]interface{})
+			if !ok {
+				return errors.New("alert it is not a map[string]interface{}")
+			}
 
 			v, err := processAlert(a)
 			if err != nil {
-				log.Println(err)
+				logger.WithError(err).Error("can not process alert")
```
gguillemas: That is very much up for discussion since there is indeed an error in processing the specific alert, but I would just "Warn" here, since we are not returning any error and it's not something that taints the whole check.

cmd/vulcan-zap/main.go
julianvilas: Done

cmd/vulcan-zap/main.go
julianvilas: Done

cmd/vulcan-zap/main.go
julianvilas: Done

PR #7, opened 2019-11-04
WIP - Snyk check

filewalkwithme:


cmd/vulcan-snyk/main.go
```diff
@@ -0,0 +1,261 @@
...
+		vulns := []report.Vulnerability{}
+		// for each pair of (snyk vuln & module name), create a vulcan vuln
+		for _, snykModulesMap := range snykVulnerabilitiesMap {
+			for moduleName, snykIssues := range snykModulesMap {
+				vulcanVulnerability := &report.Vulnerability{}
+
+				vulcanVulnerability.Summary = snykIssues[0].Title + ": " + moduleName
+				vulcanVulnerability.Description = extractOverview([]byte(snykIssues[0].Description))
+				vulcanVulnerability.Details = extractDetails([]byte(snykIssues[0].Description))
+				vulcanVulnerability.ImpactDetails = createImpactDetails(snykIssues)
```
gguillemas: I don't understand why you iterate through all of the issues here? Won't that yield the full list of dependencies that introduced issues instead of the dependencies that introduced that particular issue? Anyhow, I think impact details is meant to explain the potential impact of the vulnerability, I think this should go in resources or maybe details.

cmd/vulcan-snyk/main.go
```diff
@@ -0,0 +1,261 @@
...
+			if i < n-2 {
+				str = str + " > "
+			}
+		}
+		res = res + str + "\n"
+	}
+	return res
+}
+
+var regexpRemediationTagBegin = regexp.MustCompile(`(?i)<h2.*id="remediation".*</h2>`)
```
gguillemas: I would actually avoid parsing HTML with regex and instead find the heading in the markdown and then only convert the relevant part to HTML. I think we can live with this, but I looked into it and `markdown.Parse` returns the AST and since the markdown is simple it should be pretty easy to find the relevant node.

cmd/vulcan-snyk/main.go
```diff
@@ -0,0 +1,261 @@
...
+	bufStr = strings.ReplaceAll(bufStr, "--|", "---|")
+
+	html := markdown.ToHTML([]byte(bufStr), markdownParser, nil)
+
+	locationTagDetails := regexpDetailsTagBegin.FindIndex(html)
+	if len(locationTagDetails) > 1 {
+		html = html[locationTagDetails[1]:]
+		locationTagNextH2 := regexpNextH2TagBegin.FindIndex(html)
+		if len(locationTagNextH2) > 1 {
+			html = html[:locationTagNextH2[0]]
```
gguillemas: If you prefer using regex here, I would just capture the content between tags instead of working by index, like this:
```
var regexpOverviewTag = regexp.MustCompile(`(?i)<h2.*?id="overview">(.*?)</h2>`)
return regexpOverviewTag.FindStringSubmatch(html)[1]
```
I think that is easier to understand.

cmd/vulcan-snyk/main.go
filewalkwithme: Hi @gguillemas, Snyk will report one issue for each vulnerable library found in the dependency tree.
For example, imagine that `erubis` is imported by 9 libraries used by `Project Foo`, so Snyk will return 9 entries for that. Those entries will contain repeated information about the issue itself: description, remmediation, etc and also a `from` field, which contains specific information about which dependencies introduced that issue. 

Also, even if Snyk return multiple JSON entries for the same issue through the CLI, they are rendering a single Issue (with an aggregated `Introduced paths` field) in their UI report.

Regarding wether this information fits the `impact details` field, you are right. I have mixed `details` and `impact_details`. They are inverted. I will fix that.

cmd/vulcan-snyk/main.go
```diff
@@ -0,0 +1,298 @@
...
+		err = os.RemoveAll(repoPath)
+		if err != nil {
+			return err
+		}
+
+		if err := os.Mkdir(repoPath, 0755); err != nil {
+			return err
+		}
+
+		_, err = git.PlainClone(repoPath, false, &git.CloneOptions{
```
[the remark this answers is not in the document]
filewalkwithme: Fixed!

cmd/vulcan-snyk/main.go
gguillemas: Cool! I understand :+1:

In that case, I would really use the resources table, as each library is a resource that is affected by the vulnerability. I think a great example is the AWS Trusted Advisor check.

PR #22, opened 2019-12-16
Add the possibility of specifying score of secrets

gguillemas:


cmd/vulcan-seekret/main.go
```diff
@@ -133,19 +140,31 @@ func main() {
...
+					if score, ok := ruleScores[shortRuleName]; ok {
+						ruleScore = score.Score
+					}
+
+					if ruleScore > vuln.Score {
+						vuln.Score = ruleScore
+					}
+				}
+
+				log.Println(ruleScore)
```
julianvilas: This is not intended to stay, right?

julianvilas: I see it has been removed already.

PR #36, opened 2020-01-30
Add stream to vulcan-exposed-http-endpoint

omira-sch:


omira-sch:
OK. Added a string present on body response.

PR #52, opened 2020-03-13
Add vulcan-trivy check for scanning Docker images

danfaizer:
This PR adds `vulcan-trivy` check which is meant to scan docker images.
The build is optimised for size and pull speed by including vulnerability cache compressed within the Vulcan check docker image.

Some options are available in the check to pass as parameters to `trivy`:
- force_update_db: forces update of the vulnerability cache.
- ignore_unfixed: only show vulnerabilities that are fixed.
- severities : output only includes the severities specified.

**Limitations**: The check can only scan one private registry at the moment. This is done by comparing if the target docker image registry matches with the environment var `REGISTRY_DOMAIN`.


cmd/vulcan-trivy/vulcan-trivy.go
```diff
@@ -0,0 +1,337 @@
...
+		if score > vuln.Score {
+			vuln.Score = score
+		}
+
+		ap := map[string]string{
+			"Name":            feature.PkgName,
+			"Version":         feature.InstalledVersion,
+			"Severity":        feature.Severity,
+			"FixedBy":         feature.FixedVersion,
+			"Vulnerabilities": fmt.Sprintf("%s", feature.Description),
```
julianvilas: Now this is more the description of one vulnerability than a list of vulnerabilities, right? Maybe we can change this field to be a set of `VulnerabilityID` from the `feature`. Then if we want to keep a description because looks more cool, maybe we can add a new field.

cmd/vulcan-trivy/vulcan-trivy.go
```diff
@@ -0,0 +1,337 @@
...
+			ap["Vulnerabilities"] = fmt.Sprintf("%s ...truncated", ap["Vulnerabilities"][0:500])
+		}
+
+		// Check if affected package has already been indexed.
+		if _, exists := duppedPackageVulns[ap["Name"]]; !exists {
+			duppedPackageVulns[ap["Name"]] = ap
+			continue
+		}
+
+		key := duppedPackageVulns[ap["Name"]]
```
julianvilas: This access is being done some lines above. Maybe you can rewrite the while thing as something like:
```
elem, ok := duppedPackageVulns[ap["Name"]]
if !ok {
...
}
```

cmd/vulcan-trivy/vulcan-trivy.go
```diff
@@ -0,0 +1,337 @@
...
+		if isMoreSevere(ap["Severity"], key["Severity"]) {
+			key["Severity"] = ap["Severity"]
+			// Only showing most sever vulnerability description to avoid
+			// report size overflow.
+			if ap["Vulnerabilities"] != "" {
```
julianvilas: Taking in the consideration the recommendation from above about the `Vulnerabiltities` field, this might become and append of the `VulnerbilityID` instead.

cmd/vulcan-trivy/vulcan-trivy.go
```diff
@@ -0,0 +1,337 @@
...
+	if reportTruncated {
+		vuln.Details = generateDetails(len(rows), totalVulnerablePackages, registryEnvDomain, target)
+	}
+
+	ap.Rows = rows
+	vp.Rows = rows
+
+	vuln.Resources = append(vuln.Resources, ap, vp)
+	state.AddVulnerabilities(vuln)
+
```
julianvilas: Maybe it's good for debugging to store the report.json in the `report.Data` field.

cmd/vulcan-trivy/vulcan-trivy.go
```diff
@@ -0,0 +1,337 @@
...
+	// Build trivy command with arguments.
+	triviCmd := "./trivy"
+	triviArgs := []string{
+		"--cache-dir", trivyCachePath,
+		"-f", "json",
+		"-o", reportOutputFile,
+	}
+	// Force vulnerability db cache update.
+	if opt.ForceUpdateDB {
+		triviArgs = append(triviArgs, "--skip-update")
```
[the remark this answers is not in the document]
danfaizer: Fixed. My mistake.

cmd/vulcan-trivy/vulcan-trivy.go
```diff
@@ -0,0 +1,337 @@
...
+	}
+	// Force skip update if image is latest and force update is not explicit.
+	if !opt.ForceUpdateDB && strings.HasSuffix(target, "latest") {
+		log.Printf("Skipping cache update with a latest dockerimage: %s\n", target)
```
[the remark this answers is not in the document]
danfaizer: This was correct. I've modified to log a warning message.

cmd/vulcan-trivy/vulcan-trivy.go
```diff
@@ -0,0 +1,337 @@
...
+	}
+	// Show only vulnerabilities with specific severities.
+	if opt.Severities != "" {
+		severitiesFlag := []string{"--severity", opt.Severities}
+		triviArgs = append(triviArgs, severitiesFlag...)
+	}
+	// Append the target (docker image including registry hostname).
+	triviArgs = append(triviArgs, target)
+
+	log.Printf("Running command: %s %s\n", triviCmd, triviArgs)
```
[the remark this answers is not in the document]
danfaizer: Done. Thanks for the tip.

cmd/vulcan-trivy/vulcan-trivy.go
```diff
@@ -0,0 +1,337 @@
...
+
+	jsonFile, err := os.Open(reportOutputFile)
+	if err != nil {
+		log.Printf("trivy report output file read failed with error: %s\n", err)
+		return errors.New("trivy report output file open failed")
+	}
+	log.Printf("successfully open trivy report output file %s\n", reportOutputFile)
+	defer jsonFile.Close()
+
+	byteValue, err := ioutil.ReadAll(jsonFile)
```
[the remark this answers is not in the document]
danfaizer: Done. Thanks for the tip.

cmd/vulcan-trivy/vulcan-trivy.go
```diff
@@ -0,0 +1,337 @@
...
+			"Version",
+			"Vulnerabilities",
+		},
+	}
+
+	var rows []map[string]string
+	duppedPackageVulns := make(map[string]map[string]string)
+	// As we are scanning only one container per check, we have only one item
+	// in results array.
+	for _, feature := range results[0].Vulnerabilities {
```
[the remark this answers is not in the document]
danfaizer: Not a big deal. Changes requested applied.

cmd/vulcan-trivy/vulcan-trivy.go
danfaizer: Absolutely agree. Changed description for VulnerabilityIDs + link.
Truncated the number of VulnerabilityIDs to 10 to avoid overflow the report.

cmd/vulcan-trivy/vulcan-trivy.go
danfaizer: Good catch. Fixed.

cmd/vulcan-trivy/vulcan-trivy.go
danfaizer: Change described above. Fixed.

cmd/vulcan-trivy/vulcan-trivy.go
danfaizer: Unfortunately this is tricky. Depending on the size of the report.json the Vulcan report overflows and the check is reported as failed.

cmd/vulcan-trivy/vulcan-trivy.go
```diff
@@ -118,14 +119,14 @@ func run(ctx context.Context, target string, optJSON string, state state.State)
 		"-f", "json",
 		"-o", reportOutputFile,
 	}
-	// Force vulnerability db cache update.
-	if opt.ForceUpdateDB {
+	// Skip vulnerability db update if not explicitly forced.
+	if !opt.ForceUpdateDB {
```
julianvilas: Nitpicking but the db being updated is the default behaviour, and also you are negating this var, so why not use `opt.SkipDBUpdate` instead?

cmd/vulcan-trivy/vulcan-trivy.go
```diff
@@ -118,14 +119,14 @@ func run(ctx context.Context, target string, optJSON string, state state.State)
...
 		triviArgs = append(triviArgs, "--skip-update")
 	}
-	// Force skip update if image is latest and force update is not explicit.
+	// Log skip vulnerability db update if image is latest and force
+	// update is not explicit.
 	if !opt.ForceUpdateDB && strings.HasSuffix(target, "latest") {
```
julianvilas: This if can be embedded in the previous one:
```
if skip {
  if latest {
  }
}
```

cmd/vulcan-trivy/local.toml.example
```diff
@@ -0,0 +1,8 @@
+[Check]
+Target = "registry-1.docker.io/library/python:3.4-alpine"
+# example targets:
+# registry-1.docker.io/library/python:3.4-alpine
+# registry-1.docker.io/library/postgres:9.6
+
+# example Options
+# Options = '{"force_update_db": false, "ignore_unfixed": false, "severities":"UNKNOWN,LOW,MEDIUM,HIGH,CRITICAL"}'
```
julianvilas: Rename `force_update_db` to `skip_db_update`?

cmd/vulcan-trivy/vulcan-trivy.go
```diff
@@ -0,0 +1,332 @@
...
+
+var (
+	checkName        = "vulcan-trivy"
+	logger           = check.NewCheckLog(checkName)
+	trivyCachePath   = "trivy_cache"
+	reportOutputFile = "report.json"
+)
+
+type options struct {
+	ForceUpdateDB bool   `json:"force_update_db"`
```
julianvilas: Change to `SkipDBUpdate`?

cmd/vulcan-trivy/vulcan-trivy.go
```diff
@@ -0,0 +1,332 @@
...
+			return rows[i]["Version"] < rows[j]["Version"]
+		}
+	})
+
+	// To avoid report size overflow only top 30 vulnerabilities are shown.
+	// In addition, docker commands are provided to let the user generate
+	// the full report.
+	totalVulnerablePackages := len(rows)
+	if totalVulnerablePackages > 30 {
+		log.Print("warning: truncate to top 30 vulnerabilities\n")
```
julianvilas: Use `logger.Warn` instead?

cmd/vulcan-trivy/vulcan-trivy.go
danfaizer: As we've discussed, we want to skip vulnerability db as default behaviour.
Default value for a boolean (false) works well and you need to be explicit by setting ForceUpdateDB option if you want to update vulnerability db.
Perhaps we are making a bit confusing the logic but we are making more clear the behaviour and the effect of the options for the check.
Hopefully this has sense for you.

cmd/vulcan-trivy/vulcan-trivy.go
danfaizer: Fixed.

cmd/vulcan-trivy/local.toml.example
danfaizer: Commented above.

cmd/vulcan-trivy/vulcan-trivy.go
danfaizer: Commented above.

cmd/vulcan-trivy/vulcan-trivy.go
danfaizer: Ups 😅 , missed that one. Fixed.

PR #53, opened 2020-03-18
Only send credentials to specific GHE endpoint

gguillemas:


cmd/vulcan-seekret/main.go
```diff
@@ -79,8 +79,14 @@ func main() {
...
 
+		// TODO: Support multiple authenticated Github Enterprise instances.
+		githubURL, err := url.Parse(os.Getenv("GITHUB_ENTERPRISE_ENDPOINT"))
+		if err != nil {
+			return err
+		}
+
 		var auth *http.BasicAuth
-		if targetURL.Host != "github.com" {
+		if targetURL.Host == githubURL.Host {
```
julianvilas: Maybe we can check that the `url.Host`s are not `nil`? I don't like much relying in the behaviour of `url.Parse`, even more when we are using the `target` plain string for further steps.

I mean something like:
```
		if githubURL.Host != "" && targetURL.Host == githubURL.Host {
```

gguillemas: I actually though about this but decided against it. However, I see no harm in adding an additional check so I will push a new commit with it. Thanks!

PR #58, opened 2020-04-09
Add account alias in aws trusted advisor check

manelmontilla:
This PR modifies the aws trusted advisor check in order to add a new column to the resources table generated when findings are found for an AWS account. The value for this new column will be the same for all the rows: the first alias found for the AWS account or empty if no aliases have been found.

### DIFF
--- a/cmd/vulcan-aws-trusted-advisor/main.go
+++ b/cmd/vulcan-aws-trusted-advisor/main.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io/ioutil"
 	"net/http"
@@ -17,6 +18,7 @@ import (
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/credentials"
 	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/iam"
 	"github.com/aws/aws-sdk-go/service/support"
 	"github.com/sirupsen/logrus"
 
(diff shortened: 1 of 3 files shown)


cmd/vulcan-aws-trusted-advisor/main.go
```diff
@@ -227,11 +231,18 @@ func scanAccount(opt options, target string, logger *logrus.Entry, state state.S
 					logger.Debugf("resource with ResourceID: %s have been marked as excluded", *fr.ResourceId)
 					continue
 				}
-
-				header := []string{}
-				row := make(map[string]string)
+				// Get the alias of the account only if we did not get previously.
+				if alias == nil {
+					res, err := accountAlias(creds)
```
filewalkwithme, line 236: If `accountAlias` depends only on `creds`, why not invoke this function right after getting the credentials, outside the main loop? For example, if you could initialise the alias on line 145, no? That way you don't need to care about wether you already have initialised the `alias` variable or not.

Also, another question, if for any reason we can not retrieve the account alias, should we fail the entire check execution?

manelmontilla: Good questions thanks for making them!!
- First question: I thought about it yes but finally, decided not do it because this way if an account does not have any issue we will save a request. But now that you mention it saving a request is not that much, so maybe, make more sense to declare where you said in order to make the code cleaner. What do you think?

- Second question:  To be honest I think yes it is better to make the  check fail. I mean, I understand that having the account alias can be hardly considered essential, but, I also think that the the call for getting alias should work almost always, so I preferred to make the check fail loudly in order to be aware that this call failed. In any case I agree that if we start seeing failures of the whole check for this reason we should change this.

filewalkwithme: - Ahhhh, now I see your point! Indeed, it was smart to save a request. But also, it would be cleaner to initialise the alias outside the main loop. Now that you have clarified this to me, I would be ok with both implementations. Not strong preferences here.
- Understood and agree!

Thanks for the clarification. I will approve the PR as is, but feel free to modify it at will.

PR #60, opened 2020-05-06
Build and push

danfaizer:
Build checks, docker images and push them to Docker Hub:
https://hub.docker.com/u/vulcansec

### DIFF
--- a/_scripts/libdocker.sh
+++ b/_scripts/libdocker.sh
@@ -15,6 +15,7 @@ dkr_env() {
     cat <<"EOF"
 export DKR_USERNAME=${DKR_USERNAME:-}
 export DKR_PASSWORD=${DKR_PASSWORD:-}
+export DKR_SERVER=${DKR_SERVER:-docker.io}
 export DKR_REGISTRY=${DKR_REGISTRY:-https://registry.hub.docker.com}
 export DKR_REGISTRY_VERSION=${DKR_REGISTRY_VERSION:-v2}
 EOF
@@ -29,9 +30,9 @@ EOF
 dkr_login() {
     local -r username="${DKR_USERNAME:?docker username required}"
     local -r password="${DKR_PASSWORD:?docker password required}"
-    local -r registry="${DKR_REGISTRY:?docker registry required}"
+    local -r server="${DKR_SERVER:?docker server required}"
 
-    echo "${password}" | dkr_execute login -u "${username}" --password-stdin "${registry}"
+    echo "${password}" | dkr_execute login -u "${username}" --password-stdin "${server}"
 }
 
 ########################
--- a/_scripts/libdocker.sh
+++ b/_scripts/libdocker.sh
@@ -1,0 +1,142 @@
+#!/bin/bash
+#
+# Docker library
+
+########################
+# Load global variables used on Docker configuration
+# Globals:
+#   DKR_*
+# Arguments:
+#   None
+# Returns:
+#   Series of exports to be used as 'eval' arguments
+#########################
+dkr_env() {
+    cat <<"EOF"
+export DKR_USERNAME=${DKR_USERNAME:-}
+export DKR_PASSWORD=${DKR_PASSWORD:-}
+export DKR_REGISTRY=${DKR_REGISTRY:-https://registry.hub.docker.com}
+export DKR_REGISTRY_VERSION=${DKR_REGISTRY_VERSION:-v2}
+EOF
+}
+
+########################
+# Login to target registry.
+# Arguments:
+#   None
+# Returns:
+#   None
+dkr_login() {
+    local -r username="${DKR_USERNAME:?docker username required}"
+    local -r password="${DKR_PASSWORD:?docker password required}"
+    local -r registry="${DKR_REGISTRY:?docker registry required}"
+
+    echo "${password}" | dkr_execute login -u "${username}" --password-stdin "${registry}"
+}
+
+########################
+# Check if docker image and tag exists in target registry.
+# Arguments:
+#   $1 - Docker image name
+#   $2 - Docker image tag (Optional. Default: latest)
+# Returns:
+#   Boolean
+dkr_image_exists() {
+    local -r image_name="${1:?docker image name argument required}"
+    local -r tag="${2:-latest}"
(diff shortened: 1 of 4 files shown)


_scripts/libdocker.sh
```diff
@@ -0,0 +1,144 @@
...
+dkr_push() {
+    local -r image_name="${1:?docker image name argument required}"
+
+    if ! dkr_execute image inspect "$DKR_USERNAME/$image_name" > /dev/null; then
+        echo "Docker image [$DKR_USERNAME/$image_name] does not exist" 1>&2
+        exit 1
+    fi
+    # Ensure we are logged in
+    dkr_login
+    dkr_execute push "$DKR_USERNAME/$image_name" > /dev/null
```
jesusfcr, line 118: Add $DKR_REGISTRY to use the exact registry.

release.sh
```diff
@@ -0,0 +1,109 @@
...
+        continue
+    fi
+    # Fetch check version and modification timestamp
+    check_version=$(git_commit_id "$cf")
+    check_timestamp=$(git_timestamp "$cf")
+
+    # Check if check has been pushed due to code or dependency update
+    pushed_by_dependency=$(dkr_image_exists "$check" "dep-$dep_version")
+    if [[ $REGISTRY_RATE_LIMIT == true ]]; then sleep 1; fi
+    pushed_by_check_code=$(dkr_image_exists "$check" "$check_version")
```
jesusfcr, line 62: Do we need to take into account the dep version? this dep version shoud be explicit in the repo then in the commit.

release.sh
```diff
@@ -0,0 +1,109 @@
...
+
+    # Build docker image
+    ts_docker_build_start=$(date +"%s")
+    dkr_build "$cf" "$check"
+    ts_docker_build_finish=$(date +"%s")
+
+    # Tag docker image
+    ts_docker_tag_start=$(date +"%s")
+    dkr_tag "$check" "latest,$check_version,$check_timestamp,dep-$dep_version,dep-$dep_timestamp,$check_branch,$check_mode"
+    ts_docker_tag_finish=$(date +"%s")
```
jesusfcr, line 88: The build depends if the image is in the local docker cache or not, not in the registry.


```
if ! exists in registry then
        if ! exists in local docker  then 
             go build
             docker build
             docker tag
       end 
      if push-to-registry then
         push
     end
end
```

_scripts/libdocker.sh
```diff
@@ -0,0 +1,144 @@
...
+# Arguments:
+#   None
+# Returns:
+#   None
+dkr_login() {
+    local -r username="${DKR_USERNAME:?docker username required}"
+    local -r password="${DKR_PASSWORD:?docker password required}"
+    local -r server="${DKR_SERVER:?docker server required}"
+
+    echo "${password}" | dkr_execute login -u "${username}" --password-stdin "${server}"
```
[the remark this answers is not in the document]
danfaizer: ✅

_scripts/libdocker.sh
```diff
@@ -0,0 +1,144 @@
...
+dkr_build() {
+    local -r context="${1:?docker build context path is required}"
+    local -r image_name="${2:?docker image name argument required}"
+
+    if [ ! -d "$context" ]; then
+        echo "Context path provided [$context] does not exist or is not a directory" 1>&2
+        exit 1
+    fi
+    cd "$context" || exit
+    dkr_execute build --quiet -t "$DKR_USERNAME/$image_name" . > /dev/null
```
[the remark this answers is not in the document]
danfaizer: Done: You can now show build logs by setting:
```
DKR_PRINT_LOGS=true
```

release.sh
```diff
@@ -0,0 +1,109 @@
...
+        go_build=$((ts_go_build_finish-ts_go_build_start))
+        docker_build=$((ts_docker_build_finish-ts_docker_build_start))
+        docker_tag=$((ts_docker_tag_finish-ts_docker_tag_start))
+        docker_push=$((ts_docker_push_finish-ts_docker_push_start))
+        echo "## Build and push stats for check [$check]"
+        echo "## $total_process seconds : total process"
+        echo "## $go_build seconds : go build process"
+        echo "## $docker_build seconds : docker build process"
+        echo "## $docker_tag seconds : docker tag process"
+        echo "## $docker_push seconds : docker push process"
```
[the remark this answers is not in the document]
danfaizer: Done: I've added a one-line log line at the end of the check metrics.

release.sh
```diff
@@ -0,0 +1,109 @@
...
+
+    # Tag docker image
+    ts_docker_tag_start=$(date +"%s")
+    dkr_tag "$check" "latest,$check_version,$check_timestamp,dep-$dep_version,dep-$dep_timestamp,$check_branch,$check_mode"
+    ts_docker_tag_finish=$(date +"%s")
+
+    # Push docker image
+    ts_docker_push_start=$(date +"%s")
+    dkr_push "$check"
+    ts_docker_push_finish=$(date +"%s")
```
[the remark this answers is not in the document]
danfaizer: Done: I've refactored and added 2 env vars:
```
FORCE_BUILD="${FORCE_BUILD:-false}"
FORCE_PUSH="${FORCE_PUSH:-false}"
```
You can set the combination of them, and it will force build, force push, both or none.

_scripts/libdocker.sh
danfaizer: I'm having troubles with this one.
I will skip for now but it shouldn't be a problem if is properly defined in your local configuration.

release.sh
danfaizer: As how I'm doing the verification this is important.
For the dependency I'm taking the commit of the file (go.mod).
For the check code, I'm taking the commit of the folder.

This is in order to build only when necessary. By all means, if only one check changed but I'm taking latest commit from the repo, is like if dependencies has changed (even if not) and therefore all checks will be rebuild and pushed.
I think this is not what we want.

But I'm open to discuss via Slack.

release.sh
danfaizer: Done: I've added this to the logic.

_scripts/libdocker.sh
jesusfcr: Yes, we can keep that for another iteration

release.sh
jesusfcr: ok

PR #64, opened 2020-06-26
Add vulcan-prowler check

julianvilas:
Add a new check that uses [prowler](https://github.com/toniblyx/prowler) to determine the compliance of an AWS account against the [CIS AWS Foundations Benchmark](https://www.cisecurity.org/benchmark/amazon_web_services/).

It needs the assume role endpoint (like `vulcan-aws-trusted-advisor`) in order to get the temporary credentials to access the target account.

By default it checks against the `Level 2` profile, but can be changed passing an option.

### DIFF
--- a/cmd/vulcan-prowler/main.go
+++ b/cmd/vulcan-prowler/main.go
@@ -119,6 +119,9 @@ func main() {
 		}
 
 		endpoint := os.Getenv(envEndpoint)
+		if endpoint == "" {
+			return fmt.Errorf("%s env var must have a non-empty value", envEndpoint)
+		}
 		role := os.Getenv(envRole)
 
 		logger.Infof("using endpoint '%s' and role '%s'", endpoint, role)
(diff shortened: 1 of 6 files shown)


cmd/vulcan-prowler/main.go
```diff
@@ -0,0 +1,286 @@
...
+		if err != nil {
+			return err
+		}
+
+		endpoint := os.Getenv(envEndpoint)
+		role := os.Getenv(envRole)
+
+		logger.Infof("using endpoint '%s' and role '%s'", endpoint, role)
+
+		if err := loadCredentials(endpoint, parsedARN.AccountID, role, opts.SessionDuration); err != nil {
```
manelmontilla, line 126: I think it's not strictly required but, maybe it would be good checking that envRole and envEndpoint are at least not empty before calling loadCredentials.

julianvilas: The role is optional, but I will add check for the endpoint. Thanks!

julianvilas: Done

PR #68, opened 2020-08-05
vulcan-prowler: select group of controls according to security-level

manelmontilla:
This PR modifies the vulcan-prowler check to accept a security level for an account and set the CIS Level to check for depending on that level.
Concretely:

- If the security level is 0 or 1, it will execute the controls included in the CIS Level 1.
- If the security is not specified or it is 2, it will execute the controls included in the CIS Level 2.

Apart from that there are some small improvements on the resources table and in the descriptions of the vulnerabilities.

### DIFF
--- a/cmd/vulcan-prowler/main.go
+++ b/cmd/vulcan-prowler/main.go
@@ -9,6 +9,7 @@ import (
 	"io/ioutil"
 	"net/http"
 	"os"
+	"strings"
 
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/credentials"
@@ -44,26 +45,18 @@ var (
 	CISCompliance = report.Vulnerability{
 		Summary: "Compliance With CIS AWS Foundations Benchmark (BETA)",
 		Description: `<p>
+			The check did not received the security classification level
+			of the AWS account so the benchmark has been executed against the
+			CIS Level 2.
 			The CIS AWS Foundations Benchmark provides prescriptive
 			guidance for configuring security options for a subset of Amazon Web
 			Services with an emphasis on foundational, testable, and architecture
-			agnostic settings. The services included in the scope are:
+			agnostic settings. The services included in the scope are: IAM, Confing, 
+			CloudTrail, CloudWatch, SNS, S3 and VPC (Default).
 		</p>
 		<p>
-		<ul>
-			<li>IAM</li>
-			<li>Config</li>
-			<li>CloudTrail</li>
-			<li>CloudWatch</li>
-			<li>SNS</li>
-			<li>S3</li>
-			<li>VPC (Default)</li>
-		</ul>
-		</p>
-		<p>
-			The provided recommendations are classified in 2 different levels: 1
-			and 2, being the level 2 intended for environments or use cases where
-			security is paramount.
+			The provided recommendations are provided as a result of controls included in
+			the CIS Level 2.
 		</p>
 		<p>
 			Check the Details and Resources sections to know the compliance status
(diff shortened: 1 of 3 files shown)


cmd/vulcan-prowler/main.go
```diff
@@ -44,26 +45,47 @@ var (
 	CISCompliance = report.Vulnerability{
 		Summary: "Compliance With CIS AWS Foundations Benchmark (BETA)",
 		Description: `<p>
+			The check did not received the security classification level
```
gguillemas, line 48: `s/received/receive/`

cmd/vulcan-prowler/main.go
```diff
@@ -44,26 +45,47 @@ var (
...
+			</p>
+			<p>
+		    The CIS AWS Foundations Benchmark provides prescriptive
+			guidance for configuring security options for a subset of Amazon Web
+			Services with an emphasis on foundational, testable, and architecture
+			agnostic settings. The services included in the scope are: IAM, Confing, 
+			CloudTrail, CloudWatch, SNS, S3 and VPC (Default).
+		</p>
+		<p>
+			The provided recommendations are provided as a result controls included in
```
gguillemas, line 87: I would avoid saying provided twice in the same sentence. I also think this sentence is missing an "of".
Maybe we can leave it like: `Recommendations are provided in order to comply with all the controls required by the CIS Level 1.`

cmd/vulcan-prowler/main.go
```diff
@@ -74,13 +96,59 @@ var (
...
+			</p>
+			<p>
+		    The CIS AWS Foundations Benchmark provides prescriptive
+			guidance for configuring security options for a subset of Amazon Web
+			Services with an emphasis on foundational, testable, and architecture
+			agnostic settings. The services included in the scope are: IAM, Confing, 
+			CloudTrail, CloudWatch, SNS, S3 and VPC (Default).
+		</p>
+		<p>
+			The provided recommendations are provided as a result of controls included in
```
gguillemas, line 117: Same here as above.

cmd/vulcan-prowler/main.go
```diff
@@ -74,13 +96,59 @@ var (
...
+		References: []string{
+			"https://d0.awsstatic.com/whitepapers/compliance/AWS_CIS_Foundations_Benchmark.pdf",
+			"https://github.com/toniblyx/prowler",
+			"https://www.cisecurity.org/benchmark/amazon_web_services/",
+		},
+		Score: report.SeverityThresholdMedium,
+	}
+
+	CISComplianceInfo = report.Vulnerability{
+		Summary: "Information about CIS AWS Foundations Benchmark (BETA)",
```
gguillemas, line 133: `s/about/About/`

cmd/vulcan-prowler/main.go
```diff
@@ -74,13 +96,59 @@ var (
...
+		Description: `<p>
+			     Information gathered by executing the CIS benchmark about the account.
```
gguillemas, line 135: `s/about/on`

cmd/vulcan-prowler/main.go
```diff
@@ -74,13 +96,59 @@ var (
...
+			"https://www.cisecurity.org/benchmark/amazon_web_services/",
+		},
+		Score: report.SeverityThresholdNone,
 	}
 )
 
 type options struct {
 	Region          string   `json:"region"`
 	Groups          []string `json:"groups"`
 	SessionDuration int      `json:"sessionDuration"` // In secs.
```
gguillemas, line 150: I think we are doing `snake_case` for all the options in Vulcan. Maybe this is a good chance to change this or maybe we can do it in a separate PR.

cmd/vulcan-prowler/main.go
```diff
@@ -136,86 +204,168 @@ func main() {
...
 	}
 
 	c := check.NewCheckFromHandler(checkName, run)
 	c.RunAndServe()
 }
 
-func addVulnsToState(state state.State, r *prowlerReport, alias string) {
-	v := CISCompliance
+func groupsFromOpts(opts options) ([]string, error) {
+	// If ROLFP level is specified then it defines the group to use.
```
gguillemas, line 251: I would avoid referencing ROLFP anywhere, as this is the classification criteria that we have chosen in Adevinta. Externally I believe only the security level should be exposed, as it is independent from how each user of Vulcan actually calculates it.

cmd/vulcan-prowler/main.go
```diff
@@ -136,86 +204,168 @@ func main() {
...
+	if opts.SecurityLevel == nil {
+		return opts.Groups, nil
+	}
+	level := *opts.SecurityLevel
+	if level < 0 || level > 2 {
+		return nil, errors.New("invalid rolfp level value")
```
gguillemas, line 257: Same here with the ROLFP reference.

cmd/vulcan-prowler/main.go
gguillemas: Maybe we could also drop the "level" word like in the other descriptions.

manelmontilla: Yes thanks

cmd/vulcan-prowler/main.go
manelmontilla: Makes absolutely sense.

cmd/vulcan-prowler/main.go
manelmontilla: True thanks

cmd/vulcan-prowler/main.go
manelmontilla: Yes thanks.

cmd/vulcan-prowler/main.go
manelmontilla: 👍

cmd/vulcan-prowler/main.go
manelmontilla: Yes makes sense, I'm doing it in this PR because, as you said, we are already using it to improve a little bit the check and not only for the taking into account the security level. I'm also modifying the local.toml.example file.

cmd/vulcan-prowler/main.go
manelmontilla: Yes true.

cmd/vulcan-prowler/main.go
manelmontilla: Yes thanks

PR #69, opened 2020-08-07
Update ZAP check to work with most recent features

gguillemas:
This PR implements the following improvements to the ZAP check:

- Updates to the latest available weekly version of ZAP.
- Makes use of the new AJAX spider to better crawl single page applications.
- Supports WebAddress type assets exclusively to remove guessing the server.
- Reduce progress polling frequency to avoid spamming the persistence server.
- Fix bug where unknown CWE ID were reported as 2^32-1 overflowing the DB.

### DIFF
--- a/cmd/vulcan-zap/main.go
+++ b/cmd/vulcan-zap/main.go
@@ -72,14 +72,14 @@ func main() {
 		}
 		client, err := zap.NewClient(cfg)
 		if err != nil {
-			return fmt.Errorf("error configuring the ZAP proxy client: %v", err)
+			return fmt.Errorf("error configuring the ZAP proxy client: %w", err)
 		}
 
 		client.Core().SetOptionDefaultUserAgent("Vulcan - Security Scanner - vulcan@adevinta.com")
 
 		targetURL, err := url.Parse(target)
 		if err != nil {
-			return fmt.Errorf("error parsing target URL: %v", err)
+			return fmt.Errorf("error parsing target URL: %w", err)
 		}
 
 		if opt.Username != "" {
@@ -97,7 +97,7 @@ func main() {
 		client.Spider().SetOptionMaxDepth(opt.Depth)
 		resp, err := client.Spider().Scan(targetURL.String(), "", "", "", "")
 		if err != nil {
-			return fmt.Errorf("error executing the spider: %v", err)
+			return fmt.Errorf("error executing the spider: %w", err)
 		}
 
 		v, ok := resp["scan"]
@@ -120,7 +120,7 @@ func main() {
 			time.Sleep(10 * time.Second)
 			resp, err := client.Spider().Status(scanid)
 			if err != nil {
-				return fmt.Errorf("error getting the status of the spider: %v", err)
+				return fmt.Errorf("error getting the status of the spider: %w", err)
 			}
 
 			v, ok := resp["status"]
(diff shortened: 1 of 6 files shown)


cmd/vulcan-zap/main.go
```diff
@@ -113,21 +117,22 @@ func main() {
 		}
 
 		for {
-			time.Sleep(1 * time.Second)
+			time.Sleep(10 * time.Second)
 			resp, err := client.Spider().Status(scanid)
 			if err != nil {
-				return fmt.Errorf("error getting the status of the scan: %v", err)
+				return fmt.Errorf("error getting the status of the spider: %v", err)
```
manelmontilla, line 123: Not a big deal, because we are not going to unwrap this error anway, but I here I would use this:
fmt.Errorf("error getting the status of the spider: %w", err)

cmd/vulcan-zap/main.go
```diff
@@ -149,6 +154,39 @@ func main() {
 		logger.Debug("Waiting for spider results...")
 		time.Sleep(5 * time.Second)
 
+		logger.Debugf("Running AJAX spider %v levels deep...", opt.Depth)
+
+		client.AjaxSpider().SetOptionMaxCrawlDepth(opt.Depth)
+		resp, err = client.AjaxSpider().Scan(targetURL.String(), "", "", "")
+		if err != nil {
+			return fmt.Errorf("error executing the AJAX spider: %v", err)
```
manelmontilla: Same here I would use this, event if it is not a big deal: fmt.Errorf("error executing the AJAX spider: %w", err)

cmd/vulcan-zap/main.go
gguillemas: Ooooh, true, I did not modify the errors and they were written before wrapping was implemented in Go. I think it's a good moment to use the proper format string, so I will change it. Good catch :+1:

gguillemas: Done!

cmd/vulcan-zap/main.go
```diff
@@ -149,6 +154,39 @@ func main() {
...
+			if !ok {
+				// In this case if we can not get the status let's fail.
+				return errors.New("can not retrieve the status of the AJAX spider")
+			}
+			status, ok := v.(string)
+			if !ok {
+				return errors.New("status is present in response body when calling AjaxSpider().Scatus() but it is not a string")
+			}
+
+			if status >= "running" {
```
manelmontilla, line 182: Just a question what are the possible statuses?

cmd/vulcan-zap/main.go
```diff
@@ -149,13 +154,46 @@ func main() {
...
+				break
+			}
+		}
+
+		logger.Debug("Waiting for AJAX spider results...")
+		time.Sleep(5 * time.Second)
+
 		// Scan actively only if explicitly indicated.
 		if opt.Active {
```
manelmontilla, line 191: I would probably extract the body of this if to function, called maybe activeScan or something.
In any case if you think it is better to leave it for another PR because is not the main objective
of this PR, it is okey for me.

cmd/vulcan-zap/main.go
gguillemas: Already addressed this one in the last commit :grin:

cmd/vulcan-zap/main.go
gguillemas: Completely agree, I will do it in this same PR, I think it makes sense :ok_hand:

cmd/vulcan-zap/main.go
gguillemas: I cannot find it documented anywhere but [according to the code](https://github.com/zaproxy/zap-extensions/blob/ec25f80911d8173b11b6cf250b8f8d7b2bd57895/addOns/spiderAjax/src/main/java/org/zaproxy/zap/extension/spiderAjax/AjaxSpiderAPI.java#L82-L83) only two, so it should be fine.

cmd/vulcan-zap/main.go
gguillemas: Done!

manelmontilla (approved):
LGTM

PR #71, opened 2020-08-18
Vulcan prowler improve output

manelmontilla:
This PR modifies the vulcan-prowler-check so it:
- Receives and option called "security-level" that determines which CIS Level, 1 or 2, the check uses to assets if an AWS is compliant with.
- Generates an "improved" version of the finding generated when the AWS account has any failed a or more CIS controls.


cmd/vulcan-prowler/main.go
```diff
@@ -325,16 +363,31 @@ func fillCisLevelVuln(v *report.Vulnerability, r *prowlerReport, alias string, s
...
 		}
 	}
-
+	sort.Slice(rows, func(i, j int) bool {
+		return rows[i].score > rows[j].score
+	})
+	for _, r := range rows {
+		fcTable.Rows = append(fcTable.Rows, r.row)
+	}
 	v.Resources = append(CISCompliance.Resources, fcTable)
```
filewalkwithme, line 391: I know that this line is not part of this PR but I want to suggest a change:

Why not use the following form:
```
v.Resources = append(v.Resources, fcTable)
```

instead of:
```
v.Resources = append(CISCompliance.Resources, fcTable)
```
?

I am saying that because Golang docs for the `Append` function says that we should try to use the same variable as source and destination:

```
It is therefore necessary to store the result of append, often in the variable holding the slice itself
```
https://golang.org/pkg/builtin/#append

Again, this is just a suggestion.

cmd/vulcan-prowler/main.go
```diff
@@ -42,6 +43,8 @@ var (
 		"cislevel2",
 	}
 
+	// CISCompliance is the vulnerability generated by the check when the it
```
julianvilas, line 46: `s/when the it/when it/`

cmd/vulcan-prowler/main.go
```diff
@@ -67,9 +70,12 @@ var (
 			"https://github.com/toniblyx/prowler",
 			"https://www.cisecurity.org/benchmark/amazon_web_services/",
 		},
-		Score: report.SeverityThresholdMedium,
+		Score: report.SeverityThresholdLow,
 	}
 
+	// CISLevel1Compliance is the vulnerability generated by the check when the
```
julianvilas, line 76: Same than above

cmd/vulcan-prowler/main.go
```diff
@@ -97,9 +103,12 @@ var (
 			"https://github.com/toniblyx/prowler",
 			"https://www.cisecurity.org/benchmark/amazon_web_services/",
 		},
-		Score: report.SeverityThresholdLow,
+		Score: report.SeverityThresholdMedium,
 	}
 
+	// CISLevel2Compliance is the vulnerability generated by the check when the
```
julianvilas, line 109: Same than above

cmd/vulcan-prowler/main.go
```diff
@@ -127,9 +136,12 @@ var (
 			"https://github.com/toniblyx/prowler",
 			"https://www.cisecurity.org/benchmark/amazon_web_services/",
 		},
-		Score: report.SeverityThresholdMedium,
+		Score: report.SeverityThresholdLow,
```
julianvilas, line 139: Why the severity of the issue for level 2 is lower than for level 1?

cmd/vulcan-prowler/main.go
```diff
@@ -325,16 +363,31 @@ func fillCisLevelVuln(v *report.Vulnerability, r *prowlerReport, alias string, s
...
+			c := controlRow{row, cinfo.Severity}
+			rows = append(rows, c)
+			fallthrough
+		default:
+			total++
 		}
 	}
-
+	sort.Slice(rows, func(i, j int) bool {
+		return rows[i].score > rows[j].score
```
julianvilas, line 386: Just a silly thing, but in this case the order won't be the same in different executions. I'd put more cases for the sort, for example sorting by the control number if the score is the same.

julianvilas (changes requested):

cmd/vulcan-prowler/main.go
manelmontilla: I completely agree thanks for the suggestion!!

cmd/vulcan-prowler/main.go
manelmontilla: thanks fixed

cmd/vulcan-prowler/main.go
manelmontilla: thanks fixed

cmd/vulcan-prowler/main.go
manelmontilla: thanks fixed

cmd/vulcan-prowler/main.go
manelmontilla: That's a good question, we agreed on considering not being compliant with the CIS Level 2 not as bad as not being compliant with the CIS Level 1, which is more basic. It is true that there is no clear solution here.
What's your opinion regarding this?

cmd/vulcan-prowler/main.go
manelmontilla: That's a very good idea indeed thanks!!

cmd/vulcan-prowler/main.go
```diff
@@ -360,6 +418,8 @@ func parseControl(raw string) (control string, description string, err error) {
 	// parts[0] = [check13 .
 	control = strings.Replace(parts[0], "[check", "", -1)
 	// control = 13 .
+	control = control[:1] + "." + control[1:]
```
filewalkwithme, line 421: One last thing, if `control` is empty, this command will triggers a `slice bounds out of range` error. Can we put a protection/check for that?

manelmontilla: I think it is a good suggestion,  the prowler tool must always return a non empty string, but it is true that it is better to check for that precondition.

filewalkwithme (approved):
Thanks @manelmontilla ! LGTM

julianvilas (approved):

PR #74, opened 2020-10-09
Add vulcan-github-alerts check to report vulnerable dependencies in Github 

gguillemas:
This check uses the [Github GraphQL API](https://docs.github.com/en/free-pro-team@latest/graphql/reference/objects#securityvulnerability) to retrieve [security alerts related to vulnerable dependencies](https://docs.github.com/en/free-pro-team@latest/github/managing-security-vulnerabilities/about-alerts-for-vulnerable-dependencies) present for a repository. The check also tries to determine the minimum version necessary for each dependency to fix all the vulnerabilities reported, since the API does not currently populate the corresponding field.

Example:
image

### DIFF
--- a/cmd/vulcan-github-alerts/main.go
+++ b/cmd/vulcan-github-alerts/main.go
@@ -82,6 +82,7 @@ type ExtendedAdvisory struct {
 	References  []struct {
 		URL string `json:"url"`
 	} `json:"references"`
+	WithdrawnAt            string `json:"withdrawnAt"`
 	VulnerableVersionRange string
 	// NOTE: Github currently always returns the FirstPatchedVersion field empty.
 	FirstPatchedVersion string
@@ -151,6 +152,12 @@ func main() {
 		dependencies := map[string]*dependencyData{}
 		for _, alert := range alerts {
 			vuln := alert.SecurityVulnerability
+
+			// If the advisory has been withdrawn, we will ignore it.
+			if vuln.Advisory.WithdrawnAt != "" {
+				continue
+			}
+
 			vuln.Advisory.VulnerableVersionRange = vuln.VulnerableVersionRange
 			vuln.Advisory.FirstPatchedVersion = vuln.FirstPatchedVersion
 			advisoryScore := scoreSeverity(vuln.Advisory.Severity)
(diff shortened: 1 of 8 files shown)


kozmic:
Great check @gguillemas !

What sort of access does the `GITHUB_ENTERPRISE_TOKEN` need to have for this to run in all organisations and repositories?
Do you support private repositories?

cmd/vulcan-github-alerts/main.go
```diff
@@ -0,0 +1,331 @@
...
+
+		return nil
+	}
+
+	c := check.NewCheckFromHandler(checkName, run)
+
+	c.RunAndServe()
+}
+
+func scoreSeverity(githubSeverity string) float32 {
```
kozmic, line 280: I'm a bit uncertain of this mapping... but do not really know the best way to solve this. I would estimate 1-5% of any "vulnerable dependency" to be actually exploitable, so a lot of the `Critical` findings may not actually be critical, and actually be `Low` severity. The problem is, "what if" we get a `Critical` that is exploitable... Hopefully Github's VulnDB will get "exploitability" factor (like Snyk and Tenable), which we can use to rank findings better.

gguillemas:
> Great check @gguillemas !
> 
> What sort of access does the `GITHUB_ENTERPRISE_TOKEN` need to have for this to run in all organisations and repositories?
> Do you support private repositories?

Good question, I will add a comment on the manifest file to explain this.

Only administrators of a repository or the organization that the repository belongs to can query the security alerts related to that repository. The user that the token belongs to must be either of those two. I will do some more testing now but, as far as I know, the only permission that the token requires is `repo:admin` in order to be able to pull the alerts themselves.

One way to allow that token to test all repositories is to add the token user as an administrator-type collaborator for the repositories that wish to be able to use the check. This can also be automated for all repositories using a a Github administrator user. Doing this will allow the check to also work with private repositories.

cmd/vulcan-github-alerts/main.go
gguillemas: I agree. This is a discussion that we frequently have, since the same happens with the RetireJS check. Our position right now has been to assume the worst case, specially in the cases (such as with outdated dependencies) where resolving the issue has low impact and is recommended even if the vulnerability does not seem to be exploitable.

julianvilas (approved):

PR #76, opened 2020-10-20
Fix handling of unreachable assets error

ka3de:
This PR modifies the checks implementation to handle the error cases on which the target to be scanned is not reachable.


ka3de:
Closing in favor of implementation done in branch `handle-reachability` according to new check runner spec.

PR #79, opened 2020-11-13
Improve false positive detection mechanisms

gguillemas:
This PR improves the false positive detection mechanisms in the `vulcan-http-exposed-resources` check.

The mechanisms implemeted are:
1. Querying randomly generated resource paths to identify if the server returns false `200 OK` statuses.
2. Performing the exact same check twice to identify if the server returns `200 OK` responses inconsistently.
3. Validating if the rate at which the server responds `200 OK` to requests for exposed resources is abnormally high.

Additionally, a new informational vulnerability will be added to the report in the first case in order to keep a record of the server behaviour in the case of false negatives. All false positives are now logged in the details section and, if false positives are identified, the vulnerability is reported with the highest score among the high confidence findings.

### DIFF
--- a/cmd/vulcan-exposed-http-resources/main.go
+++ b/cmd/vulcan-exposed-http-resources/main.go
@@ -90,7 +90,7 @@ var (
 
 	// falseOKSuffixes are the suffixes that will be added to the randomly
 	// generated resource to check if the server is returning false OK responses.
-	falseOKSuffixes = [...]string{"", "/", ".txt", ".html", ".php", ".asp", ".jsp"}
+	falseOKSuffixes = [...]string{"", "/", "/images/", ".txt", ".html", ".php", ".asp", ".jsp"}
 
 	// falsePositivesMessage is the message that will be displayed
 	// in the details section if the exposed resources of the check
--- a/cmd/vulcan-exposed-http-resources/main.go
+++ b/cmd/vulcan-exposed-http-resources/main.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"context"
+	"crypto/rand"
 	"crypto/tls"
 	"encoding/json"
 	"fmt"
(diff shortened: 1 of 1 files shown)


cmd/vulcan-exposed-http-resources/main.go
```diff
@@ -234,6 +289,7 @@ func checkResource(targetURL *url.URL, httpResource Resource) []map[string]strin
 					return
 				}
 
+				// This line creates a copy of the net.URL object.
 				targetResource, err := url.Parse(targetURL.String())
```
manelmontilla, line 293: Not a big deal, but you don't need to parse the url again to make a copy, you can do something like this:
https://play.golang.org/p/H-2gGbNwKsh
But to be honest there is no difference in terms of performance or anything.

cmd/vulcan-exposed-http-resources/main.go
```diff
@@ -339,46 +395,116 @@ func checkPath(targetResource *url.URL, httpResource Resource) (bool, error) {
...
+	rand.Read(bytes)
+	token := fmt.Sprintf("%x", bytes)
+
+	falseOK := false
+	resources := []map[string]string{}
+
+	// We request the non-existent resource with various suffixes.
+	for _, suffix := range falseOKSuffixes {
+		// This line creates a copy of the net.URL object.
+		targetResource, err := url.Parse(targetURL.String())
```
manelmontilla, line 415: Same as commented before regarding alternative way of doing a  copy of the url

manelmontilla (approved):
LGTM

cmd/vulcan-exposed-http-resources/main.go
gguillemas: I was not aware of this, it is a good thing to know for the future! In this case, I think I will leave it like it is since it was not introduced in this PR and I like that it takes a single line that is easily understandable. Thanks a lot for the tip!

PR #81, opened 2020-12-04
Improve ZAP to make it production-ready

gguillemas:
This PR implementes the following improvements to the `vulcan-zap` check:
- Fixes a bug where vulnerabilities would be scored only on a scale from 0 to 3.
- Fixes a deadlock when the ZAP daemon died and the check kept polling forever.
- Implements an option to filter out vulnerabilities below a minimum score.
- Log crawled URLs in both the regular spider and the AJAX spider.
- Allow disabling ZAP scanners through the options.
- Explicitly set the target scope.


danfaizer (approved):
Very good job @gguillemas 
The check has improved a lot.
👍

PR #83, opened 2020-12-18
Increase polling interval for nessus check

ka3de:
This PR modifies nessus check in order to:

- Increase the default polling interval from 30s to 5min
- Set an initial delay time before starting communication with Tenable API.

This is in order to try to mitigate spikes and overloading of Tenable API during scans.


jesusfcr (approved):
LGTM

PR #84, opened 2020-12-30
Avoid negatives in CWEID for uint32 type

danfaizer:
zap `cweid` default value is `-1`.
In order to avoid integer overflow when converting from zap `alert` to vulcan report `vulnerability`, in case `cweid`  is negative or bigger than int max, we set value to `0`.


filewalkwithme (approved):
LGTM

PR #85, opened 2021-01-04
Improve AWS trusted advisor checks refreshing

ka3de:
This PR improves the logic put in place for the vulcan-aws-trusted-advisor when some of its checks are enqueued for processing so we should wait up until the defined refresh TO option polling their status to verify if they are refreshed.
Also for this purpose it modifies the default refresh timeout from 5 to 60 seconds.

### DIFF
--- a/cmd/vulcan-aws-trusted-advisor/main.go
+++ b/cmd/vulcan-aws-trusted-advisor/main.go
@@ -157,6 +157,7 @@ func scanAccount(opt options, target, assetType string, logger *logrus.Entry, st
 	// poll it's status and wait up until opt.RefreshTimeout
 	if enqueued > 0 {
 		t := time.NewTicker(time.Duration(opt.RefreshTimeout) * time.Second)
+		defer t.Stop()
 
 	LOOP:
 		for {
@@ -190,7 +191,6 @@ func scanAccount(opt options, target, assetType string, logger *logrus.Entry, st
 				time.Sleep(rfrshInterval)
 			}
 		}
-		t.Stop()
 	}
 
 	// Retrieve checks summaries
--- a/cmd/vulcan-aws-trusted-advisor/main.go
+++ b/cmd/vulcan-aws-trusted-advisor/main.go
@@ -153,42 +153,45 @@ func scanAccount(opt options, target, assetType string, logger *logrus.Entry, st
 		}
 	}
 
-	// Start polling for checks refresh status
-	t := time.NewTicker(time.Duration(opt.RefreshTimeout) * time.Second)
+	// If some check was enqueued for refreshing
+	// poll it's status and wait up until opt.RefreshTimeout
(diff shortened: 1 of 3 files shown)


cmd/vulcan-aws-trusted-advisor/main.go
```diff
@@ -148,12 +152,48 @@ func scanAccount(opt options, target, assetType string, logger *logrus.Entry, st
...
+			default:
+				checkStatus, err := s.DescribeTrustedAdvisorCheckRefreshStatuses(
+					&support.DescribeTrustedAdvisorCheckRefreshStatusesInput{
+						CheckIds: checkIds,
+					},
+				)
+				if err != nil {
+					if awsErr, ok := err.(awserr.Error); ok {
+						if awsErr.Code() != "InvalidParameterValueException" {
+							return err
```
filewalkwithme, line 175: If we exit here the `time.Ticker` declared on line 159 will continue running?

filewalkwithme (changes requested):
Hi @ka3de, one question, if I understood correctly, this PR is improving the process of waiting for all Trusted Advisor checks to be refreshed before start collecting results.

However, it seems that if we reach the timeout of `60` seconds, then the program will proceed anyway, even if there are checks that were not refreshed yet.

Is this the intended behaviour?

cmd/vulcan-aws-trusted-advisor/main.go
ka3de: If we exit here the ticker will be "deallocated" no? Because we exit the context on which the ticker was declared, no?

ka3de:
> Hi @ka3de, one question, if I understood correctly, this PR is improving the process of waiting for all Trusted Advisor checks to be refreshed before start collecting results.
> 
> However, it seems that if we reach the timeout of `60` seconds, then the program will proceed anyway, even if there are checks that were not refreshed yet.
> 
> Is this the intended behaviour?

Hi @filewalkwithme, yes it's the intended behavior. The idea is to not block the check execution or wait forever to let the checks refresh, we just want to set a better upper bound than before and wait up until that time for the checks to be refreshed, if they are not refreshed during that period of time, then we proceed with the execution whatever the sate is.

cmd/vulcan-aws-trusted-advisor/main.go
ka3de: Change it for a defer stop of the ticker right after initialization, it won't be stopped explicitly after the LOOP, but it will before exiting the function which should guarantee we don't leak it.

filewalkwithme: Perfect! Thanks, @ka3de!

filewalkwithme (approved):
@ka3de Great, makes sense. I am approving the PR now. 

LGTM

PR #93, opened 2021-03-25
Fix counterproductive panics in retirejs

ka3de:
This PR modifies vulcan-retirejs check to avoid panicking on purpose for unnecessary reasons during check handle execution.
Panicking in this situations for permanent errors makes vulcan-agent handle the check execution in an incorrect way, which instead of marking it as FAILED, returns the message to the queue and makes this eventually be processed `MaxRead` times by the agent with the same execution process and result for each time.

Instead, the correct behavior implemented here is to return an error from within check handle execution so the vulcan-agent can parse it correctly and set the check status to FAILED on its first execution.


gguillemas (approved):
LGTM

PR #94, opened 2021-03-26
Fix retirejs cmd execution args

ka3de:
This PR fixes vulcan-retirejs check execution, more specifically the arguments used to call the underlying NPM _retire_ command.


gguillemas (approved):
LGTM

PR #105, opened 2021-06-09
Include details of the non-existent bucket in s3 subdomain takeover vulnerability

julianvilas:
Currently the vulnerability doesn't reflect how actually the information was found and the name of the missing bucket. This PR adds that information inside the vulnerabilty.Details field so the user can understand and reproduce the issue.

image


manelmontilla (approved):
LGTM

PR #107, opened 2021-09-08
Update drupal vulnerabilities

alexanderkjall:
Two changes here:

* Removed the generic recommendation section because that it was outdated and instead reuse the recommendations from the original security notices.
* Added all the missing notices from 2019 and forward from https://www.drupal.org/security .


kozmic:
Maybe phrase these `If you are using Drupal 8 , upgrade to Drupal 8.4.5` as `If you are using Drupal 8 , upgrade to Drupal 8.4.5 (or newer)`.

julianvilas (approved):

PR #111, opened 2021-09-08
Gitignore improvements

alexanderkjall:
Deleted a checked in binary file that I'm guessing shouldn't be there `cmd/vulcan-aws-trusted-advisor/vulcan-aws-trusted-advisor`.

Added gitignore rules for all binaries generated by the checks.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,6 @@
 *.swp
 local.toml
+
+# editor files
+.idea/
+vulcan-checks.iml
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ local.toml
 # editor files
 .idea/
 vulcan-checks.iml
+.vscode/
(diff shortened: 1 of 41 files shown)


.gitignore
```diff
@@ -1,2 +1,6 @@
 *.swp
 local.toml
+
+# editor files
+.idea/
+vulcan-checks.iml
```
julianvilas, line 6: Could you and a newline at the end of the file? Helpful when running bash commands/scripts :)

julianvilas (changes requested):
Just a minor comment, for the rest looks super good to me :)

.gitignore
```diff
@@ -1,2 +1,6 @@
 *.swp
 local.toml
+
+# editor files
```
jesusfcr, line 4: And this too just in case:

`.vscode/`

.gitignore
alexanderkjall: fixed :)

.gitignore
alexanderkjall: Added :)

julianvilas (approved):

PR #119, opened 2021-10-11
Expand vulcan-nessus vulnerabilities and add labels

julianvilas:
* Expands the resources (i.e. affected ports) of each Nessus vulnerability into individual vulnerabilities, so they can be individually actioned. If a port is not specified by Nessus, use the whole target as the affected resource.
* Calculates the fingerprint of each vulnerability so actions applied to them can be dismissed if the context changes (e.g. the output of the Nessus plugin has changed)
* Adds labels to the vulnerabilities, mostly to indicate that they are issues because the check has done some dynamic tests to check it
* Use the new CVSS v3 Nessus score if available, falling back to the already used CVSS base score.
* Ensure that all the `solution` and `see_also` references provided by Nessus are added to the vulnerability. Before we were adding just the first collection of them.

### DIFF
--- a/cmd/vulcan-nessus/nessus.go
+++ b/cmd/vulcan-nessus/nessus.go
@@ -6,7 +6,6 @@ package main
 
 import (
 	"context"
-	"crypto/sha256"
 	"encoding/json"
 	"errors"
 	"fmt"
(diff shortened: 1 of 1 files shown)


julianvilas:
Totally agree. I'll try to detail a bit more how the check works.

PR #122, opened 2021-10-18
Expand vulcan-vulners into different findings

gianz74:
- Expands the resources of each into individual vulnerabilities, so they can be individually actioned.
- Calculates the fingerprint of each vulnerability so actions applied to them can be dismissed if the context changes (e.g. new CVEs are published for the specific version of software)


cmd/vulcan-vulners/main.go
```diff
@@ -265,20 +278,28 @@ func analyzeReport(target string, nmapReport *gonmap.NmapRun) ([]report.Vulnerab
...
+				for _, row := range f.Resources.Rows {
+					cves = append(cves, row["CVE"])
+				}
+				v, ok := uniqueVulns[uniqueVulnId]
 				if !ok {
 					v.Vuln = report.Vulnerability{
-						Summary:         summary,
-						Description:     fmt.Sprintf(vulnersVuln.Description, port.Service.Product),
-						Recommendations: vulnersVuln.Recommendations,
+						ID:               computeVulnerabilityID(cves),
```
julianvilas, line 290: We added a new field to store the fingerprint of the vulnerability that is being calculated here. The field is called `Fingerprint`, so we need to use it instead of the `ID` attribute.

Apart from that, I'd suggest also adding the `Score` of the CVEs, because if one of the CVEs is re-scored we might want to reopen the issue and let the user decide again.

julianvilas (changes requested):
Thanks a lot for the great contribution @gianz74, and sorry for the late reply.

The PR looks awesome, I'm adding just a few comments according to some changes we applied since you opened the PR.

cmd/vulcan-vulners/main.go
```diff
@@ -265,20 +278,28 @@ func analyzeReport(target string, nmapReport *gonmap.NmapRun) ([]report.Vulnerab
...
+						Summary:          summary,
+						Description:      fmt.Sprintf(vulnersVuln.Description, port.Service.Product),
+						Recommendations:  vulnersVuln.Recommendations,
+						AffectedResource: fmt.Sprintf("%s, %d/%s", string(cpe), port.PortId, port.Protocol),
```
julianvilas, line 294: Oh, I also forgot to mention in the review that we should add also the `Labels` attribute to the finding, specifying at least the type of finding. In the case of `vulcan-vulners` we would be using the label `issue` for all the findings. So it could be something like:
```
Labels: []string{"issue"},
```

cmd/vulcan-vulners/main.go
```diff
@@ -306,22 +327,32 @@ func analyzeReport(target string, nmapReport *gonmap.NmapRun) ([]report.Vulnerab
...
-					Description:     fmt.Sprintf(vulnersVuln.Description, port.Service.Product),
-					Score:           f.Score,
-					Recommendations: vulnersVuln.Recommendations,
+
+					ID:               computeVulnerabilityID(cves),
+					Summary:          summary,
+					Description:      fmt.Sprintf(vulnersVuln.Description, port.Service.Product),
+					Score:            f.Score,
+					Recommendations:  vulnersVuln.Recommendations,
+					AffectedResource: fmt.Sprintf("%s, %d/%s", productID, port.PortId, port.Protocol),
```
julianvilas, line 347: Oh, I also forgot to mention in the review that we should add also the `Labels` attribute to the finding, specifying at least the type of finding. In the case of `vulcan-vulners` we would be using the label `issue` for all the findings. So it could be something like:
```
Labels: []string{"issue"},
```

cmd/vulcan-vulners/main.go
```diff
@@ -241,6 +242,18 @@ func findingByProdVers(s, v, t string) (*vulnersFinding, error) {
...
+
+	sort.Strings(elems)
+
+	for _, e := range elems {
+		fmt.Fprintf(h, "%s - ", e)
+	}
+
+	return fmt.Sprintf("%x", h.Sum(nil))
+}
+
```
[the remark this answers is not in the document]
gianz74: done, thanks.

cmd/vulcan-vulners/main_test.go
```diff
@@ -0,0 +1,27 @@
...
+
+	for _, tc := range tt {
+		t.Run(fmt.Sprintf("%s", tc.ok), func(t *testing.T) {
+			n := computeVulnerabilityID(tc.elements)
+			if tc.ok != n {
+				t.Fatalf("expected result to be %s; got %s", tc.ok, n)
+			}
+		})
+	}
+}
```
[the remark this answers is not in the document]
gianz74: removed.

julianvilas (approved):
LSGTM! Thanks for the contribution :)

PR #130, opened 2021-12-15
Remove default branch in seekret

jesusfcr:
- Removes default master branch. To allow the check to use the default ref served by the git server (master, main , ...)
- Use a fixed path for the repo, preventing the issue when the git url doesn't have a path (i.e. http://localhost:port/)

### DIFF
--- a/cmd/vulcan-seekret/main.go
+++ b/cmd/vulcan-seekret/main.go
@@ -130,11 +130,7 @@ func main() {
 		if opt.Branch != "" {
 			co.ReferenceName = plumbing.ReferenceName(path.Join("refs/heads", opt.Branch))
 		}
-		_, err = git.PlainClone(repoPath, false, &git.CloneOptions{
-			URL:   target,
-			Auth:  auth,
-			Depth: opt.Depth,
-		})
+		_, err = git.PlainClone(repoPath, false, &co)
 		if err != nil {
 			return err
 		}
--- a/cmd/vulcan-seekret/main.go
+++ b/cmd/vulcan-seekret/main.go
@@ -30,7 +30,7 @@ import (
 )
 
 type options struct {
-	Depth int `json:"depth"`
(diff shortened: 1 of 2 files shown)


cmd/vulcan-seekret/main.go
```diff
@@ -117,16 +117,23 @@ func main() {
...
 
+		co := git.CloneOptions{
+			URL:   target,
+			Auth:  auth,
+			Depth: opt.Depth,
+		}
+		if opt.Branch != "" {
+			co.ReferenceName = plumbing.ReferenceName(path.Join("refs/heads", opt.Branch))
+		}
 		_, err = git.PlainClone(repoPath, false, &git.CloneOptions{
```
manelmontilla, line 133: I think you intended to use here the CloneOptions created above in the line: 125, so I guees you should modify theese lines by:
```go
_, err = git.PlainClone(repoPath, &co)
```

cmd/vulcan-seekret/manifest.toml
```diff
@@ -2,4 +2,4 @@ Description = "Finds leaked secrets in a Git repository using Seekret"
 Timeout = 600
 AssetTypes = ["GitRepository"]
 RequiredVars = ["GITHUB_ENTERPRISE_ENDPOINT", "GITHUB_ENTERPRISE_TOKEN"]
-Options = '{"depth": 1, "branch": "master"}'
+Options = '{"depth": 1, "branch": ""}'
```
manelmontilla, line 5: I maybe put here a comment stating that, if the branch option is empty,  the check will scan the default branch of the repo.

manelmontilla (changes requested):

cmd/vulcan-seekret/main.go
jesusfcr: Done, thanks!

manelmontilla (approved):
LGTM

PR #137, opened 2022-01-21
Use created context in spider and scan

danfaizer:
This PR fixes the context misuse.
In fact, the context is created at the beginning of the process but is not passed to the spider actions nor to the scan action and therefore the only URL/Path scanned were the ones resolved by the [recurse option](https://www.zaproxy.org/docs/api/#ascanactionscan) and the target URL provided.
I would appreciate a careful review on the changes.


gguillemas (approved):
LGTM

PR #139, opened 2022-01-28
Add additional zap options

jesusfcr:
Add new options for Vulcan-zap:

- max_spider_duration: Applies to both normal an Ajax spiders.
- max_scan_duration: Limits the total amount of time spent in all the rules.
- max_rule_duration: Limits the time spent in one rule.
- openapi_url: Allows to specify a Openapi/swagger url.
- openapi_host: Used with the previous one.

### DIFF
--- a/cmd/vulcan-zap/main.go
+++ b/cmd/vulcan-zap/main.go
@@ -170,16 +170,17 @@ func main() {
 			return fmt.Errorf("error disabling scanners for passive scan: %w", err)
 		}
 
-		logger.Printf("Running spider %v levels deep...", opt.Depth)
-
 		client.Spider().SetOptionMaxDepth(opt.Depth)
 		if err != nil {
-			return fmt.Errorf("error setting max depth: %w", err)
+			return fmt.Errorf("error setting spider max depth: %w", err)
 		}
 		client.Spider().SetOptionMaxDuration(opt.MaxSpiderDuration)
 		if err != nil {
-			return fmt.Errorf("error setting max duration: %w", err)
+			return fmt.Errorf("error setting spider max duration: %w", err)
 		}
+
+		logger.Printf("Running spider %v levels deep, max duration %v, ...", opt.Depth, opt.MaxSpiderDuration)
+
 		resp, err := client.Spider().Scan(targetURL.String(), "", contextName, "", "")
 		if err != nil {
 			return fmt.Errorf("error executing the spider: %w", err)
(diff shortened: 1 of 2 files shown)


cmd/vulcan-zap/manifest.toml
```diff
@@ -6,4 +6,14 @@ Timeout = 36000 # 10 hours. Expressed in seconds as an integer.
...
 # Source: https://www.zaproxy.org/docs/alerts/
-Options = '{"depth": 2, "active": true, "min_score": 0, "disabled_scanners": ["10062", "10003", "10108"]}'
+Options = """{
+    "depth": 2, 
+    "active": true, 
+    "min_score": 0, 
+    "disabled_scanners": ["10062", "10003", "10108"],
+    "max_spider_duration": 0,
+    "max_scan_duration": 0, 
+    "max_rule_duration": 0,
```
jesusfcr, line 16: Should we set some defaults congruent with the default check timeout of 10 h, i.e. max_scan_duration: 500 to obtain the results before the check is timed out?

cmd/vulcan-zap/main.go
```diff
@@ -153,14 +158,29 @@ func main() {
...
+
 		_, err = client.Pscan().DisableScanners(disabledScanners)
 		if err != nil {
 			return fmt.Errorf("error disabling scanners for passive scan: %w", err)
 		}
 
-		logger.Printf("Running spider %v levels deep...", opt.Depth)
-
 		client.Spider().SetOptionMaxDepth(opt.Depth)
+		if err != nil {
```
danfaizer, line 174: Before, we weren't checking the error. I think it is better indeed to check the error however, then you need to read it by adding `_, err =` before the call to the function, otherwise you are reusing "the latest error value".

cmd/vulcan-zap/main.go
```diff
@@ -153,14 +158,29 @@ func main() {
...
+			return fmt.Errorf("error setting spider max depth: %w", err)
+		}
+		client.Spider().SetOptionMaxDuration(opt.MaxSpiderDuration)
```
danfaizer, line 177: Same as explained above.

cmd/vulcan-zap/main.go
```diff
@@ -232,10 +252,15 @@ func main() {
 		}
 		logger.Printf("Spider found the following URLs: %+v", resp)
 
-		logger.Printf("Running AJAX spider %v levels deep...", opt.Depth)
+		client.AjaxSpider().SetOptionMaxDuration(opt.MaxSpiderDuration)
```
danfaizer, line 255: Same as explained above.

cmd/vulcan-zap/manifest.toml
danfaizer: In my opinion, until we don't define a convention within the team (if to indicate all the available options in the manifest or not) you can leave the options with the "most convenient" default values.

Just to confirm, `0` value for `duration` parameters means infinite, right ?

If that is the case, I won't change `duration` options to match check timeout in order to keep the same behaviour we have at the moment.

danfaizer (changes requested):
Thanks for the contribution!
I've added some comments before proceed with the approval.

cmd/vulcan-zap/main.go
jesusfcr: Absolutely, thanks!

cmd/vulcan-zap/manifest.toml
jesusfcr: Makes sense, thanks

danfaizer (approved):
LGTM

PR #141, opened 2022-02-02
Expand vulcan-prowler

julianvilas:
* Findings are not expanded because it's very hard to retrieve the affected resource from Prowler output due to its format.
* Each vulnerability is labeled as `compliance`.
* No fingerprint is provided for now, because we don't expect people to mark the reported things as false positives. We might revisit this in the future.


cmd/vulcan-prowler/main.go
```diff
@@ -70,12 +70,14 @@ var (
 			Check the Details and Resources sections to know the compliance status
 			and more details.
 		</p>`,
+		Labels: []string{"compliance", "cis"},
```
danfaizer, line 73: I would suggest to add `aws` tag in this and following `Labels` entries.
As better tagged are findings better filtering we can apply later.
However, it is true that you can infer that is an AWS resource from the target/assettype.

danfaizer (dismissed):
I've added a suggestion, but feel free to dismiss if you think is not adding value.
👍

cmd/vulcan-prowler/main.go
julianvilas: You are totally right, thx u :)

It's done now!

danfaizer (approved):

PR #142, opened 2022-02-02
Expand vulcan-exposed-rdp vulnerabilities and add labels

mm-sch:
- Adapt to the affected resources model
- Calculates the fingerprint of each vulnerability so actions applied to them can be dismissed if the context changes
- Adds labels to the vulnerabilities

### DIFF
--- a/cmd/vulcan-exposed-rdp/main.go
+++ b/cmd/vulcan-exposed-rdp/main.go
@@ -7,6 +7,7 @@ package main
 import (
 	"bytes"
 	"context"
+	"crypto/sha256"
 	"encoding/binary"
 	"encoding/json"
 	"fmt"
@@ -121,10 +122,24 @@ func isExposedRDP(host, port string, timeout int) []report.Vulnerability {
 
 	if CCPDURegex.Match(b) && b[5] == byte('\xd0') {
 		exposedRDP.Details += fmt.Sprintf("RDP host detected in port: %s", port)
+		exposedRDP.AffectedResource = fmt.Sprintf("%s", port)
+		exposedRDP.ID = computeVulnerabilityID(host, exposedRDP.AffectedResource, port)
+		exposedRDP.Labels = []string{"issue", "rdp"}
 		return []report.Vulnerability{exposedRDP}
 	}
 
 	return nil
+}
+
+func computeVulnerabilityID(target, affectedResource string, elems ...interface{}) string {
+	h := sha256.New()
+
+	fmt.Fprintf(h, "%s - %s", target, affectedResource)
+
+	for _, e := range elems {
+		fmt.Fprintf(h, " - %v", e)
+	}
+	return fmt.Sprintf("%x", h.Sum(nil))
 }
 
 func run(ctx context.Context, target, assetType, optJSON string, state checkstate.State) (err error) {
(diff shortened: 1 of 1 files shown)


cmd/vulcan-exposed-rdp/main.go
```diff
@@ -121,12 +122,26 @@ func isExposedRDP(host, port string, timeout int) []report.Vulnerability {
...
+		exposedRDP.AffectedResource = fmt.Sprintf("%s", port)
+		exposedRDP.ID = computeVulnerabilityID(host, exposedRDP.AffectedResource, port)
+		exposedRDP.Labels = []string{"issue", "rdp"}
 		return []report.Vulnerability{exposedRDP}
 	}
 
 	return nil
 }
 
+func computeVulnerabilityID(target, affectedResource string, elems ...interface{}) string {
```
julianvilas, line 134: This function is now available from the SDK as: ```helpers.ComputeFingerprint(elems ...interface{}) string```

cmd/vulcan-exposed-rdp/main.go
```diff
@@ -121,12 +122,26 @@ func isExposedRDP(host, port string, timeout int) []report.Vulnerability {
 
 	if CCPDURegex.Match(b) && b[5] == byte('\xd0') {
 		exposedRDP.Details += fmt.Sprintf("RDP host detected in port: %s", port)
+		exposedRDP.AffectedResource = fmt.Sprintf("%s", port)
```
julianvilas, line 125: Should we add the protocol too? Just to make it clearer for the user.

cmd/vulcan-exposed-rdp/main.go
```diff
@@ -121,12 +122,26 @@ func isExposedRDP(host, port string, timeout int) []report.Vulnerability {
...
+		exposedRDP.ID = computeVulnerabilityID(host, exposedRDP.AffectedResource, port)
```
julianvilas, line 126: Now there's a new `Vulnerability.Fingerprint` attributed to be used instead of the `ID`. Also we are not adding the target nor the affected resource to calculate the fingerprint now.

Suggested change:
```
v.Fingerprint = helpers.ComputeFingerprint()
```
(Adding the port it's harmless and we can keep it if you want)

julianvilas (changes requested):
Hey @mm-sch, thx for the contribution. It looks very good! Just a few minor comments related to changes we applied since you did it :)

cmd/vulcan-exposed-rdp/main.go
mm-sch: Agree, I'll add it to the affected resource for more clarity.

cmd/vulcan-exposed-rdp/main.go
mm-sch: Okey! Using the `Fingerprint` attribute now, also removed all args in the fingerprint calculation.

cmd/vulcan-exposed-rdp/main.go
mm-sch: Alright, using the SDK function now.

cmd/vulcan-exposed-rdp/main.go
```diff
@@ -122,26 +121,15 @@ func isExposedRDP(host, port string, timeout int) []report.Vulnerability {
 
 	if CCPDURegex.Match(b) && b[5] == byte('\xd0') {
 		exposedRDP.Details += fmt.Sprintf("RDP host detected in port: %s", port)
-		exposedRDP.AffectedResource = fmt.Sprintf("%s", port)
-		exposedRDP.ID = computeVulnerabilityID(host, exposedRDP.AffectedResource, port)
+		exposedRDP.AffectedResource = fmt.Sprintf("%s/rdp", port)
```
julianvilas, line 124: Should the `rdp` be `tcp` instead?

julianvilas (commented):
Just a clarification more :)

cmd/vulcan-exposed-rdp/main.go
mm-sch: Done!

julianvilas (approved):
LGTM! Thx for the contribution :)

PR #147, opened 2022-02-08
Expand tls resources into different findings

julianvilas:
* Each failure detected by cipherscan will be reported as a separated Vulnerability
* The fingerprint takes into account contextual things that may change
* All the vulnerabilities will be labeled as issues


danfaizer (approved):
Very smooth! Like the result!
👍

PR #148, opened 2022-02-09
Expand exposed-hdfs resources into vulnerabilities

julianvilas:
* The affected resource is the port and protocol
* The labels are `issue` or `informational` depending on whether the finding has been confirmed or not
* The fingerprint is the product service and version detected by Nmap, and the value of the confirmed field

To test it you can use this vulnerable target:
```
docker run -it --rm -d --name hadoop-docker \
        -p 2122:2122 \
        -p 8030-8088:8030-8088 \
        -p 19888:19888 \
        -p 49707:49707 \
        -p 50010-50090:50010-50090 \
        sequenceiq/hadoop-docker /etc/bootstrap.sh -bash
```


danfaizer (approved):
LGTM

PR #150, opened 2022-02-11
Expand vulcan-certinfo vulnerabilities and add labels 

danfaizer:
- Moved vulnerability link comments to references and improved some comment style.
- AffectedResource is defined by the certificate fingerprint. AffectedResourceString is populated with the target for readability reasons. Fingerprint is added but with the "empty" value as the AffectedResource definition per se if a fingerprint.
- Vulnerabilities will be labeled as issues
- Fixed issue when checking "certificate has expired or is not yet valid" error


cmd/vulcan-certinfo/main.go
```diff
@@ -216,6 +218,9 @@ func main() {
 		}
 
 		for _, certInfo := range checker.certificateInfo {
+			certInfo.AffectedResource = checker.fingerprint
+			certInfo.AffectedResourceString = target
```
julianvilas, line 222: How would you see using the `opt.Port` instead?

julianvilas (commented):
Looks very good! Great job. Just added a suggestion to be used as the `AffectedResourceString` instead of the target.

cmd/vulcan-certinfo/main.go
danfaizer: Thanks for the review @julianvilas 
Agree, as we've discussed, it is convenient to add `opt.Port` in the AffectedResource and AffectedResourceString attributes.
Done here [720fa7b](https://github.com/adevinta/vulcan-checks/pull/150/commits/720fa7b458e09317e79bde705c0dbc4e5f722ed2)

julianvilas: Thx!

PR #153, opened 2022-02-14
Migrate vulcan-spf to v2

julianvilas:
* Vulnerabilities are labeled as issues, except those that are informational
* The affected resource is the target itself
* Default fingerprint provided


danfaizer (approved):
👍

PR #154, opened 2022-02-15
Migrate vulcan-host-discovery to v2

mmmgili:
- Discovered hosts are now splitted in different vulnerabilities.
- The affected resource is the discovered host IP.
- The fingerprint is computed using the host IP and hostname.
- Vulnerabilities are labeled as "issue" and "discovery".

### DIFF
--- a/cmd/vulcan-host-discovery/main.go
+++ b/cmd/vulcan-host-discovery/main.go
@@ -148,7 +148,7 @@ nmapReportLoop:
 			exposedVuln.AffectedResource = networkResource["IP Address"]
 			exposedVuln.Resources = []report.ResourcesGroup{gr}
 			exposedVuln.Fingerprint = helpers.ComputeFingerprint(networkResource["IP Address"], networkResource["Hostname"])
-			exposedVuln.Labels = []string{"potential"}
+			exposedVuln.Labels = []string{"issue", "discovery"}
 			vs = append(vs, exposedVuln)
 		}
 	}
--- a/cmd/vulcan-host-discovery/main.go
+++ b/cmd/vulcan-host-discovery/main.go
@@ -14,6 +14,7 @@ import (
 	gonmap "github.com/lair-framework/go-nmap"
 
 	check "github.com/adevinta/vulcan-check-sdk"
+	"github.com/adevinta/vulcan-check-sdk/helpers"
 	"github.com/adevinta/vulcan-check-sdk/helpers/nmap"
 	"github.com/adevinta/vulcan-check-sdk/state"
 	report "github.com/adevinta/vulcan-report"
@@ -72,8 +73,9 @@ func isHostInKnownHosts(host string, knownHosts []string) bool {
 }
 
 func discoveredHosts(targetSubnet string, knownHosts []string, nmapReport *gonmap.NmapRun) []report.Vulnerability {
+	var vs []report.Vulnerability
 	gr := report.ResourcesGroup{
-		Name: "Network Resources",
+		Name: "Network Resource",
 		Header: []string{
 			"Subnet",
 			"IP Address",
@@ -87,13 +89,13 @@ func discoveredHosts(targetSubnet string, knownHosts []string, nmapReport *gonma
 	 * summary to "Discovered Hosts" instead of "Unknown Hosts".
 	 */
 	if len(knownHosts) == 0 {
-		exposedVuln.Summary = "Discovered Hosts"
-		exposedVuln.Description = "At least one host has been discovered in the network"
+		exposedVuln.Summary = "Discovered Host"
+		exposedVuln.Description = "A host has been discovered in the network"
 		exposedVuln.Recommendations = []string{
-			"Identify that all of the discovered hosts are known hosts",
+			"Identify if the discovered host is a known host",
 		}
 		exposedVuln.Score = report.SeverityThresholdNone
-		exposedVuln.ImpactDetails = "If any of the discovered hosts is not known, it may be an unauthorized malicious host."
+		exposedVuln.ImpactDetails = "If the discovered host is not known, it may be an unauthorized malicious host."
 	}
 
 	add := false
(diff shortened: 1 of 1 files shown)


cmd/vulcan-host-discovery/main.go
```diff
@@ -87,13 +89,13 @@ func discoveredHosts(targetSubnet string, knownHosts []string, nmapReport *gonma
 	 * summary to "Discovered Hosts" instead of "Unknown Hosts".
 	 */
 	if len(knownHosts) == 0 {
-		exposedVuln.Summary = "Discovered Hosts"
-		exposedVuln.Description = "At least one host has been discovered in the network"
+		exposedVuln.Summary = "Discovered Host"
+		exposedVuln.Description = "A host has been discovered in the network"
```
julianvilas, line 93: In this case I think it shouldn't be a big deal because probably there won't be open issues but, in the case they were, the issue is determined by the `Summary` and `Description` fields, so if you change those a new issue would be created. Therefore we would loose the real age of the associated findings.

If you want we can check together if there are open issues in the Vulnerability DB or not.

cmd/vulcan-host-discovery/main.go
```diff
@@ -140,13 +142,17 @@ nmapReportLoop:
...
-
-	if add {
-		exposedVuln.Resources = append(exposedVuln.Resources, gr)
-		return []report.Vulnerability{exposedVuln}
+		gr.Rows = []map[string]string{networkResource}
+
+		if add {
+			exposedVuln.AffectedResource = networkResource["IP Address"]
+			exposedVuln.Resources = []report.ResourcesGroup{gr}
+			exposedVuln.Fingerprint = helpers.ComputeFingerprint(networkResource["IP Address"], networkResource["Hostname"])
```
julianvilas, line 150: The IP is already part of the affected resource, so if it changes we'll get a new different finding. The fingerprint is meant to reflect the context of the current finding, so we shouldn't include nor the target nor content present in the affected resource field. The `Hostname` or nothing are good candidates for me.

cmd/vulcan-host-discovery/main.go
```diff
@@ -140,13 +142,17 @@ nmapReportLoop:
...
+			exposedVuln.Labels = []string{"issue", "discovery"}
+			vs = append(vs, exposedVuln)
+		}
 	}
+	return vs
 
 	return nil
```
julianvilas, line 157: I think we can remove this due to the logic changes you applied.

cmd/vulcan-host-discovery/main.go
```diff
@@ -87,13 +89,13 @@ func discoveredHosts(targetSubnet string, knownHosts []string, nmapReport *gonma
...
 		exposedVuln.Recommendations = []string{
-			"Identify that all of the discovered hosts are known hosts",
+			"Identify if the discovered host is a known host",
 		}
 		exposedVuln.Score = report.SeverityThresholdNone
-		exposedVuln.ImpactDetails = "If any of the discovered hosts is not known, it may be an unauthorized malicious host."
+		exposedVuln.ImpactDetails = "If the discovered host is not known, it may be an unauthorized malicious host."
 	}
 
 	add := false
```
julianvilas, line 101: Due to the changes in the logic you applied, I think we won't need to use this anymore.

Also we could replace the content on [lines 137-138](https://github.com/adevinta/vulcan-checks/pull/154/files#diff-29023f4b1137a4edd3c3a579d05cd02401b38499937ba0e09f6329f5d76e2e34R137-R138) by:
```
vuln := exposedVuln
```
And from there make references to that `vuln` instead. Otherwise at the end of the loop all your entries in the array might end having the same values.

cmd/vulcan-host-discovery/main.go
```diff
@@ -140,13 +142,17 @@ nmapReportLoop:
...
 		}
-		gr.Rows = append(gr.Rows, networkResource)
-	}
-
-	if add {
-		exposedVuln.Resources = append(exposedVuln.Resources, gr)
-		return []report.Vulnerability{exposedVuln}
+		gr.Rows = []map[string]string{networkResource}
+
+		if add {
```
julianvilas, line 147: We can remove this too.

cmd/vulcan-host-discovery/main.go
```diff
@@ -140,13 +142,17 @@ nmapReportLoop:
...
+			exposedVuln.AffectedResource = networkResource["IP Address"]
+			exposedVuln.Resources = []report.ResourcesGroup{gr}
+			exposedVuln.Fingerprint = helpers.ComputeFingerprint(networkResource["IP Address"], networkResource["Hostname"])
+			exposedVuln.Labels = []string{"issue", "discovery"}
+			vs = append(vs, exposedVuln)
```
julianvilas, line 152: As I mentioned above, use `vuln` instead.

julianvilas (changes requested):
Great job Marc! Thx for the contribution :) 

I'd suggest some minor changes to be applied.

cmd/vulcan-host-discovery/main.go
```diff
@@ -132,23 +132,24 @@ nmapReportLoop:
...
 			"Subnet":     targetSubnet,
 			"IP Address": strings.Join(hostIPAddresses, ","),
 			"Hostname":   strings.Join(hostHostnames, ","),
 		}
-		gr.Rows = append(gr.Rows, networkResource)
-	}
+		gr.Rows = []map[string]string{networkResource}
+
+		vuln.AffectedResource = networkResource["IP Address"]
+		vuln.Resources = []report.ResourcesGroup{gr}
```
julianvilas, line 145: I think we still have the problem here, because we are reusing the same `ResourcesGroup` in all the rounds of the loop. We should move the creation of the `gr` inside the loop (from L77 to L137, for example).

cmd/vulcan-host-discovery/main.go
```diff
@@ -132,23 +132,24 @@ nmapReportLoop:
...
+		vuln.Fingerprint = helpers.ComputeFingerprint(networkResource["Hostname"])
+		vuln.Labels = []string{"issue", "discovery"}
+		vs = append(vs, vuln)
 
-	if add {
-		exposedVuln.Resources = append(exposedVuln.Resources, gr)
-		return []report.Vulnerability{exposedVuln}
 	}
+	return vs
```
julianvilas, line 152: Remove empty line.

julianvilas (changes requested):
Just a couple of final comments, for the rest LGTM!

cmd/vulcan-host-discovery/main.go
julianvilas: My bad, in this case there's no problem by doing it this way, sorry :)

julianvilas (dismissed):

julianvilas (approved):

PR #155, opened 2022-02-15
Migrate exposed-http to v2

jesusfcr:
- Create an issue per exposed port
- The affected resource is the host:part
- Labeled as `issue` and protocolo (i.e. `tcp`)

### DIFF
--- a/cmd/vulcan-exposed-http/main.go
+++ b/cmd/vulcan-exposed-http/main.go
@@ -89,9 +89,9 @@ func exposedHTTP(target string, nmapReport *gonmap.NmapRun, state checkstate.Sta
 				continue
 			}
 			v := report.Vulnerability{
-				AffectedResource: fmt.Sprintf("%s:%d", target, port.PortId),
+				AffectedResource: fmt.Sprintf("%d/%s", port.PortId, port.Protocol),
 				Labels:           []string{port.Protocol, "issue"},
-				Fingerprint:      helpers.ComputeFingerprint(port),
+				Fingerprint:      helpers.ComputeFingerprint(port.Service.Product, port.Service.Version),
 				Summary:          "Exposed HTTP Port",
 				Description:      "An HTTP server is listening at least in one port ot the server.",
 				Score:            report.SeverityThresholdNone,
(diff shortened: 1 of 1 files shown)


cmd/vulcan-exposed-http/main.go
```diff
@@ -79,56 +80,50 @@ var (
...
-				"Protocol": port.Protocol,
-				"Service":  port.Service.Product,
-				"Version":  port.Service.Version,
-			}
-			if port.Service.Tunnel == "ssl" {
-				networkResource["SSL"] = "yes"
-			}
-			gr.Rows = append(gr.Rows, networkResource)
+			v := report.Vulnerability{
+				AffectedResource: fmt.Sprintf("%s:%d", target, port.PortId),
```
julianvilas, line 92: The target is already an entity of the finding, so it's not necessary to add it again in the affected resource. If you check the current Live Report UI you'll see that after you expand the target, you see the affected resources for that target, so it would look a bit redundant to see the target also there.

I'd suggest to use `port/protocol` instead, like for example `80/tcp`.

cmd/vulcan-exposed-http/main.go
```diff
@@ -79,56 +80,50 @@ var (
...
+				Labels:           []string{port.Protocol, "issue"},
+				Fingerprint:      helpers.ComputeFingerprint(port),
```
julianvilas, line 94: The port is already part of the affected resource, so it the port changes we have a new different finding. The fingerprint is meant to reflect the context of the current finding, so we shouldn't include nor the target nor content present in the affected resource field.

I'd suggest using the `Product` and `Version` fields, because if you acknowledge something related to a specific thing in that port, it should be dismissed if the thing behind that port changes to a different thing.

cmd/vulcan-exposed-http/main.go
```diff
@@ -79,56 +80,50 @@ var (
...
+						"Version",
+						"SSL",
+					},
+					Rows: []map[string]string{{
+						"Hostname": target,
+						"Port":     strconv.Itoa(port.PortId),
+						"Protocol": port.Protocol,
+						"Service":  port.Service.Product,
+						"Version":  port.Service.Version,
+						"SSL": func() string {
```
julianvilas, line 114: Good one! :)

cmd/vulcan-exposed-http/main.go
```diff
@@ -79,56 +80,50 @@ var (
...
-			}
-			gr.Rows = append(gr.Rows, networkResource)
+			v := report.Vulnerability{
+				AffectedResource: fmt.Sprintf("%s:%d", target, port.PortId),
+				Labels:           []string{port.Protocol, "issue"},
+				Fingerprint:      helpers.ComputeFingerprint(port),
+				Summary:          "Exposed HTTP Port",
+				Description:      "An HTTP server is listening at least in one port ot the server.",
+				Score:            report.SeverityThresholdNone,
+				Resources: []report.ResourcesGroup{{
```
julianvilas, line 98: This is totally a matter of taste, but I'd separate the `{` on different lines, like:
```
Resources: []report.ResourcesGroup{
    {
    ...
    },
}
```

julianvilas (changes requested):
Very good job! thx for the contribution :)

Just adding some small comments.

cmd/vulcan-exposed-http/main.go
jesusfcr: Makes sense.

cmd/vulcan-exposed-http/main.go
jesusfcr: Got it!

cmd/vulcan-exposed-http/main.go
jesusfcr: :)

cmd/vulcan-exposed-http/main.go
jesusfcr: Sure

julianvilas (approved):

PR #156, opened 2022-02-15
Migrate aws-alerts to v2

jesusfcr:


cmd/vulcan-aws-alerts/main.go
```diff
@@ -70,8 +70,8 @@ func getCredentials(url string, accountID, role string) (*credentials.Credential
 	if role != "" {
 		m["role"] = role
 	}
-	jsonBody, err := json.Marshal(m)
-	req, err := http.NewRequest("POST", url, bytes.NewBuffer(jsonBody))
+	jsonBody, _ := json.Marshal(m)
+	req, _ := http.NewRequest("POST", url, bytes.NewBuffer(jsonBody))
```
julianvilas, line 74: I guess you are removing the `err` because we were not checking it, but then I'd suggest checking the errors instead.

cmd/vulcan-aws-alerts/rds.go
```diff
@@ -77,28 +61,41 @@ func caCertificateRotation(target string, vulcanAssumeRoleEndpoint string, roleN
...
+									Labels:           []string{"issue"},
+									Fingerprint:      helpers.ComputeFingerprint(details.AutoAppliedAfterDate),
+									Summary:          `Managed AWS databases using CA about to expire`,
+									Score:            report.SeverityThresholdHigh,
+									Description: `Due to the expiration of the AWS RDS CA, and to prevent downtime ` +
+										`in your applications, you should add the new CA to your clients using a ` +
+										`managed (i.e. RDS or Aurora) database through SSL/TLS and perform maintenance ` +
+										`on the affected database instances before the certificate expiration date.`,
+									References: []string{"https://aws.amazon.com/blogs/database/amazon-rds-customers-update-your-ssl-tls-certificates-by-february-5-2020/"},
+									Resources: []report.ResourcesGroup{{
```
julianvilas, line 76: This is totally a matter of taste, but I'd separate the `{` on different lines, like:
```
Resources: []report.ResourcesGroup{
    {
    ...
    },
}
```

julianvilas (commented):
Very good job Jesus! :)

I'm adding just a couple of non-blocking cosmetic comments.

cmd/vulcan-aws-alerts/main.go
jesusfcr: Ok, I didn't do because the input parameters are well known and it's unlikely that happend. But agree it's easy and better to do it. Thanks!

cmd/vulcan-aws-alerts/rds.go
jesusfcr: Sure

julianvilas (approved):

PR #167, opened 2022-03-03
Migrate vulcan-exposed-bgp to v2

julianvilas:
* Change some attributes of the vulnerability
* Use the port as the affected resource
* Labelled as informational
* Computing a fingerprint without args


jesusfcr (approved):
LGTM

PR #168, opened 2022-03-03
Migrate vulcan-mx to v2

julianvilas:
* Small refactor of the check
* Use the target as affected resource
* Labelled as informational
* Using a fingerprint with no args


jesusfcr (approved):
LGTM

PR #169, opened 2022-03-04
Expand smtp-open-relay resources into vulns

jesusfcr:
Expand resources into vulnerabilities

Validated against:
```sh
docker run -p 587:25 -p 25:25 -d -e POSTFIX_myhostname=smtp.domain.tld mwader/postfix-relay
```

### DIFF
--- a/cmd/vulcan-smtp-open-relay/main.go
+++ b/cmd/vulcan-smtp-open-relay/main.go
@@ -13,6 +13,7 @@ package main
 import (
 	"context"
 	"encoding/json"
+	"fmt"
 	"strconv"
 	"strings"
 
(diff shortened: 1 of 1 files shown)


cmd/vulcan-smtp-open-relay/main.go
```diff
@@ -62,46 +63,39 @@ var (
...
-				networkResource := map[string]string{
-					"Hostname": target,
-					"Port":     strconv.Itoa(port.PortId),
-				}
-				gr.Rows = append(gr.Rows, networkResource)
+				vuln := openRelay
+				vuln.AffectedResource = fmt.Sprintf("%d/%s", port.PortId, port.Protocol)
+				vuln.Fingerprint = helpers.ComputeFingerprint()
+				vuln.Labels = []string{"issue", "smtp", "discovery"}
+				vuln.Resources = []report.ResourcesGroup{{
```
jesusfcr, line 82: Do we need to keep this? as it is redundant.

jesusfcr:
Should we set as critical?

cmd/vulcan-smtp-open-relay/main.go
julianvilas: You mean the `Resources` right? If so I totally agree. Kill it with fire!

julianvilas:
> Should we set as critical?

That is a good question. I'm checking for example [how Tenable reports it](https://www.tenable.com/plugins/nessus/10262) and they report as `Critical` for `CVSS v2` and `High` for `CVSS v3`. I'm a bit more on the side of reporting it as `High` because the exploitation impact is limited. For me `Critical` means full compromise.

julianvilas (approved):
LGTM!

cmd/vulcan-smtp-open-relay/main.go
jesusfcr: Done

julianvilas (approved):

PR #170, opened 2022-03-04
Expand exposed-router-ports resources into vulns

jesusfcr:
- Change some attributes of the vulnerability
- Use the port / protocol as the affected resource
- Labelled as informational
- Computing a fingerprint with service.program

### DIFF
--- a/cmd/vulcan-exposed-router-ports/main.go
+++ b/cmd/vulcan-exposed-router-ports/main.go
@@ -7,6 +7,7 @@ package main
 import (
 	"context"
 	"encoding/json"
+	"fmt"
 	"strconv"
 	"strings"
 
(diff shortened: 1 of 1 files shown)


cmd/vulcan-exposed-router-ports/main.go
```diff
@@ -61,44 +62,36 @@ var (
...
-				"Port":     strconv.Itoa(port.PortId),
-				"Protocol": port.Protocol,
-				"Service":  port.Service.Product,
-				"Version":  port.Service.Version,
-			}
-			gr.Rows = append(gr.Rows, networkResource)
+			vuln := exposedVuln
+			vuln.AffectedResource = fmt.Sprintf("%d/%s", port.PortId, port.Protocol)
+			vuln.Fingerprint = helpers.ComputeFingerprint(port.Protocol, port.Service)
+			vuln.Labels = []string{"issue", port.Protocol, "discovery"}
```
julianvilas, line 74: I'd avoid adding dynamic labels for now, as the current approach is that labels are stored at _Issue_ level in the Vulnerability DB. There are ongoing discussions about it but we don't know yet what will be the outcome of those.

cmd/vulcan-exposed-router-ports/main.go
```diff
@@ -61,44 +62,36 @@ var (
...
-				"Hostname": target,
-				"Port":     strconv.Itoa(port.PortId),
-				"Protocol": port.Protocol,
-				"Service":  port.Service.Product,
-				"Version":  port.Service.Version,
-			}
-			gr.Rows = append(gr.Rows, networkResource)
+			vuln := exposedVuln
+			vuln.AffectedResource = fmt.Sprintf("%d/%s", port.PortId, port.Protocol)
+			vuln.Fingerprint = helpers.ComputeFingerprint(port.Protocol, port.Service)
```
julianvilas, line 73: I'd add the `Product` instead of the `Protocol`. In fact the second doesn't provide value to the fingerprint because when the protocol changes, the affected resource changes too, and therefore will be actually a different finding.

julianvilas (changes requested):
Just two minor comments :)

cmd/vulcan-exposed-router-ports/main.go
julianvilas: Also, in this case the label should be `informational` instead of `issue`.

jesusfcr: Ok, so? just remove it?
And, is it `issue` or `informational`?

jesusfcr: ok

cmd/vulcan-exposed-router-ports/main.go
jesusfcr: Gotcha!

julianvilas (approved):

PR #172, opened 2022-03-04
Expand exposed http endpoint

manelmontilla:
~~This PR modifies  the check: ``vulcan-exposed-http-endpoints`` so:~~
~~* It fixes a bug that was preventing the check to test all the default rules.~~
~~* Expands the vulnerabilities it generates, so they comply with v2 model. Concretely the check generates a vulnerability per each url tested, even if the same URL matches multiples rules, only one vulnerability is generated with the field ``AfectedResource``  set to the URL.~~
~~For calculating the fingerprint, all the rules matching the url reported in a vulnerability are sorted and used to generated the fingerprint. That means: the vulnerability will be considered modified if the definitions of the rules affecting a URL have changed, or the set of rules matching that URL have changed.~~

This PR deprecates the `vulcan-exposed-http-endpoint` in favour of `vulcan-exposed-http-resources`.


cmd/vulcan-exposed-http-endpoint/main.go
```diff
@@ -45,54 +45,73 @@ If no paths are specified the check will test all the paths present in _paths/*.
...
+		return nil, nil
 	}
-	if len(resources) > 0 {
-		vuln = &exposedVuln
-		vuln.Resources[0].Rows = resources
+	var vulns []report.Vulnerability = make([]report.Vulnerability, 0, len(vulnerablePaths))
+	for url, paths := range vulnerablePaths {
+		v := exposedVuln
+		v.AffectedResource = url
+		v.Fingerprint = fingerprint(paths)
```
danfaizer, line 94: I think that the fingerprint should include `the context` in which the vulnerability has been found.
For example, in this check I would recommend to use only options `Status` and `RegExp`.
We have the following hierarchy:
Target -> AffectedResource -> Fingerprint
Making use of the AffectedResource as Fingerprint in this case would be redundant and won't make us able to detect if the finding conditions (status or regexp) has changed in order to detect the vulnerability.

cmd/vulcan-exposed-http-endpoint/main.go
```diff
@@ -45,54 +45,73 @@ If no paths are specified the check will test all the paths present in _paths/*.
...
+		vulns = append(vulns, v)
 	}
-	return vuln, nil
+	return &vulns, nil
+}
+
+func fingerprint(paths []path.Path) string {
```
danfaizer, line 100: As we are creating a vulnerability per path and in relation with the explanation above I think you can just use `helpers.ComputeFingerprint` SDK function to compute the fingerprint.
I am right or I'm missing something?

danfaizer (changes requested):
Thanks for the contribution @manelmontilla
I've added some comments. If I misunderstood something or you want some clarification about the comments feel free to reach me 😊

cmd/vulcan-exposed-http-endpoint/main.go
```diff
@@ -45,54 +45,73 @@ If no paths are specified the check will test all the paths present in _paths/*.
...
+		vulnerablePaths[url] = matches
+	}
+	if len(vulnerablePaths) == 0 {
+		return nil, nil
 	}
-	if len(resources) > 0 {
-		vuln = &exposedVuln
-		vuln.Resources[0].Rows = resources
+	var vulns []report.Vulnerability = make([]report.Vulnerability, 0, len(vulnerablePaths))
+	for url, paths := range vulnerablePaths {
```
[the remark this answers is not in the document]
julianvilas: As we discussed, I just deprecated the check.

cmd/vulcan-exposed-http-endpoint/main.go
```diff
@@ -45,54 +45,73 @@ If no paths are specified the check will test all the paths present in _paths/*.
...
+		v := exposedVuln
+		v.AffectedResource = url
```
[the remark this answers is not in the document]
julianvilas: As we discussed, I just deprecated the check.

cmd/vulcan-exposed-http-endpoint/main.go
julianvilas: As we discussed, I just deprecated the check.

cmd/vulcan-exposed-http-endpoint/main.go
julianvilas: As we discussed, I just deprecated the check.

danfaizer (approved):

PR #143, opened 2022-02-02
Expand vulcan-exposed-db vulnerabilities and add labels

mm-sch:
- Expands the Exposed DB vulnerability's resources table into individual vulnerabilities, so they can be individually actioned
- Calculates the fingerprint of each vulnerability so actions applied to them can be dismissed if the context changes
- Adds labels to the vulnerabilities

PR #174, opened 2022-03-08
Migrate vulcan-exposed-services to v2 

gguillemas:
- Expands each exposed service into an its own vulnerability so it can be individually actioned.
- Calculates the fingerprint of each vulnerability so actions applied to them can be dismissed if the context changes.
- Adds labels to the vulnerabilities.

Almost identical to #143.


danfaizer (approved):
LGTM

PR #177, opened 2022-03-21
Migrate vulcan-masscan to v2

julianvilas:
* Expand each open port to a different finding
* Labelled as informational
* Computing a fingerprint with no args


danfaizer (approved):
👍

PR #178, opened 2022-03-28
Upgrade trivy

jesusfcr:
- Upgrade trivy base image to 0.24.4
- Adapt `--download-db-only` breaking change.
- Leverage `TRIVY_CACHE_DIR` env variable
- Add information about the context (os packages, lang-packages, ...)
- Clearly identify the path where the reference is (i.e. path to a go binary, path to a jar file). 
- Fill the resource table with the vulnerabilities.

### DIFF
--- a/cmd/vulcan-trivy/vulcan-trivy.go
+++ b/cmd/vulcan-trivy/vulcan-trivy.go
@@ -242,7 +242,7 @@ func processVulns(results ScanResponse, registryEnvDomain, target string, state 
 	vp := report.ResourcesGroup{
 		Name: "Package Vulnerabilities",
 		Header: []string{
-			"FixedBy",
+			"Fixed Version",
 			"Vulnerabilities",
 			"Severity",
 			"CWEs",
@@ -272,7 +272,7 @@ func processVulns(results ScanResponse, registryEnvDomain, target string, state 
 				maxScore = newScore
 			}
 			row := make(map[string]string, len(vp.Header))
-			row["FixedBy"] = p.fixedBy
+			row["Fixed Version"] = p.fixedBy
 			if p.cwes != nil {
 				urls := []string{}
 				for _, cwe := range p.cwes {
@@ -292,15 +292,15 @@ func processVulns(results ScanResponse, registryEnvDomain, target string, state 
 			Name: "Packages",
 			Header: []string{
 				"Location",
-				"FixedBy",
+				"Min. Recommended Version",
 			},
 			Rows: []map[string]string{},
 		}
 		for path := range det.paths {
 			prg.Rows = append(prg.Rows,
 				map[string]string{
-					"Location": path,
-					"FixedBy":  det.fixedBy,
+					prg.Header[0]: path,
+					prg.Header[1]: det.fixedBy,
 				})
 		}
 
--- a/cmd/vulcan-trivy/vulcan-trivy.go
+++ b/cmd/vulcan-trivy/vulcan-trivy.go
@@ -262,18 +262,20 @@ func processVulns(results scanResponse, registryEnvDomain, target string, state 
 
 		vp.Rows = []map[string]string{}
 		maxScore := getScore("NONE")
-		fingerprint := make([]string, len(l))
+		fingerprint := []string{}
 		for i, p := range l {
 			fingerprint = append(fingerprint, p.cve+p.severity)
+
 			// Compute the fingerprint for all the CVEs but add only vulnCVETruncateLimit to the table
-			if i > vulnCVETruncateLimit {
+			if i >= vulnCVETruncateLimit {
 				continue
 			}
+
 			newScore := getScore(p.severity)
 			if newScore > maxScore {
 				maxScore = newScore
 			}
-			row := make(map[string]string, len(vp.Header))
+			row := map[string]string{}
 			row["Fixed Version"] = p.fixedBy
 			if p.cwes != nil {
 				urls := []string{}
@@ -367,6 +369,7 @@ func getScore(severity string) float32 {
 
 var cveRegex = regexp.MustCompile(`^CVE-(\d{4})-(\d+)$`)
 
+// cve2num returns a numeric representation with year and id in case of CVE or a 0 otherwise
 func cve2num(cve string) int {
 	m := cveRegex.FindStringSubmatch(cve)
 	if len(m) == 3 {
--- a/cmd/vulcan-trivy/vulcan-trivy.go
+++ b/cmd/vulcan-trivy/vulcan-trivy.go
@@ -201,7 +201,8 @@ func processVulns(results scanResponse, registryEnvDomain, target string, state 
 			path := ""
 			switch {
 			case tt.Class == "os-pkgs":
-				path = tt.Type // alpine, centos, ...
+				// Type contains the os distro name (i.e. alpine, centos, amazon, ...)
+				path = fmt.Sprintf("%s:%s", tt.Type, tv.PkgName)
 			case tv.PkgPath != "":
 				path = tv.PkgPath
 			default:
@@ -301,13 +302,13 @@ func processVulns(results scanResponse, registryEnvDomain, target string, state 
 			Details: generateDetails(registryEnvDomain, target),
 			Resources: []report.ResourcesGroup{
 				{
-					Name: "Packages",
+					Name: "Package",
 					Header: []string{
-						"Location",
+						"Package",
 						"Min. Recommended Version",
 					},
 					Rows: []map[string]string{{
-						"Location":                 key.path,
+						"Package":                  key.path,
 						"Min. Recommended Version": det.fixedBy,
 					}},
 				},
--- a/cmd/vulcan-trivy/vulcan-trivy.go
+++ b/cmd/vulcan-trivy/vulcan-trivy.go
@@ -151,6 +151,8 @@ func run(ctx context.Context, target, assetType, optJSON string, state checkstat
 		severitiesFlag := []string{"--severity", opt.Severities}
 		triviArgs = append(triviArgs, severitiesFlag...)
 	}
+	// Restrict to vulnerabilities (no config/secrets yet)
+	triviArgs = append(triviArgs, "--security-checks", "vuln")
 	// Append the target (docker image including registry hostname).
 	triviArgs = append(triviArgs, target)
 
--- a/cmd/vulcan-trivy/vulcan-trivy.go
+++ b/cmd/vulcan-trivy/vulcan-trivy.go
@@ -17,6 +17,7 @@ import (
 	"time"
 
 	report "github.com/adevinta/vulcan-report"
+	"github.com/mcuadros/go-version"
 
 	check "github.com/adevinta/vulcan-check-sdk"
 	"github.com/adevinta/vulcan-check-sdk/helpers"
@@ -24,7 +25,7 @@ import (
 	"github.com/avast/retry-go"
 )
 
-const vulnCVETrucateLimit = 30
+const vulnCVETrucateLimit = 10
 
 var (
 	checkName        = "vulcan-trivy"
@@ -63,10 +64,14 @@ type ScanResponse []struct {
 
 type outdatedKey struct {
 	name    string
-	path    string
 	version string
 }
 
+type vulnData struct {
+	packages []outdatedPackage
+	paths    map[string]interface{}
+	fixedBy  string
+}
 type outdatedPackage struct {
 	severity string
 	fixedBy  string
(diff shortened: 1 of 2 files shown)


cmd/vulcan-trivy/vulcan-trivy.go
```diff
@@ -186,116 +194,108 @@ func run(ctx context.Context, target, assetType, optJSON string, state checkstat
...
-			Labels: []string{"potential", "docker"},
-
-			// Finding attributes.
-			Fingerprint:      fingerprint,
-			AffectedResource: affectedResource,
-			Score:            getScore(op.severity),
-			Details:          generateDetails(registryEnvDomain, target),
-			Resources:        []report.ResourcesGroup{vp},
-		}
+			vuln.Fingerprint = helpers.ComputeFingerprint()
```
jesusfcr, line 282: What would be the fingerprint?

cmd/vulcan-trivy/vulcan-trivy.go
```diff
@@ -186,116 +194,108 @@ func run(ctx context.Context, target, assetType, optJSON string, state checkstat
...
+				// Target includes the image uri (i.e. adevinta/vulcan-local:latest (alpine 3.15.1)) -> Removed
+				// Path is empty -> removed
+				// Example "os-pkgs alpine libretls:3.3.4-r2"
+				vuln.AffectedResource = strings.TrimSpace(fmt.Sprintf("%s %s %s:%s", tt.Class, tt.Type, key.name, key.version))
+			} else if key.path != "" {
+				// Example: "lang-pkgs jar Java-flyway/lib/aad/jackson-databind-2.10.5.1.jar com.fasterxml.jackson.core:jackson-databind:2.10.5.1"
+				vuln.AffectedResource = strings.TrimSpace(fmt.Sprintf("%s %s %s-%s %s:%s", tt.Class, tt.Type, tt.Target, key.path, key.name, key.version))
+			} else {
+				// Example: "lang-pkgs gobinary app/vulcan-api github.com/docker/distribution:v2.7.1+incompatible"
+				vuln.AffectedResource = strings.TrimSpace(fmt.Sprintf("%s %s %s %s:%s", tt.Class, tt.Type, tt.Target, key.name, key.version))
```
jesusfcr, line 294: Review the format of the affectedResource.

cmd/vulcan-trivy/vulcan-trivy.go
```diff
@@ -186,116 +194,108 @@ func run(ctx context.Context, target, assetType, optJSON string, state checkstat
...
-			return opvArr[i].name < opvArr[j].name
-		default:
-			return opvArr[i].version < opvArr[j].version
+		for key, l := range outdatedPackageVulns {
+			// Sort outdated packages by severity
+			sort.Slice(l, func(i, j int) bool {
+				return getScore(l[i].severity) > getScore(l[j].severity)
+			})
+			// Keep only top vulnTruncateLimit
+			if len(l) > vulnTruncateLimit {
```
jesusfcr, line 225: Do we have to limit it?

cmd/vulcan-trivy/vulcan-trivy.go
```diff
@@ -186,116 +194,108 @@ func run(ctx context.Context, target, assetType, optJSON string, state checkstat
...
+					row["CWEs"] = strings.Join(l.cves, ",")
+				}
+				row["CVE"] = l.cve
+				if l.link == "" {
+					l.link = fmt.Sprintf("https://nvd.nist.gov/vuln/detail/%s", l.cve)
+				}
+				row["Link"] = l.link
+				row["Severity"] = l.severity
+				vp.Rows = append(vp.Rows, row)
+			}
```
jesusfcr, line 265: Should we keep this or compute from the from CWES? What if more than one?

cmd/vulcan-trivy/vulcan-trivy.go
```diff
@@ -186,116 +191,102 @@ func run(ctx context.Context, target, assetType, optJSON string, state checkstat
...
-					severity: dockerVuln.Severity,
-					fixedBy:  dockerVuln.FixedVersion,
-					cves:     []string{},
+				row := make(map[string]string, len(vp.Header))
+				row["FixedBy"] = l.fixedBy
+				if l.cves != nil {
+					row["CWEs"] = strings.Join(l.cves, ",")
 				}
+				row["CVE"] = l.cve
+				if l.link == "" {
```
jesusfcr, line 249: Does it make sense to use the trivy provided link vs the nist.gov one?

- https://avd.aquasec.com/nvd/2021/cve-2021-21707/
- https://nvd.nist.gov/vuln/detail/cve-2021-21707

cmd/vulcan-trivy/vulcan-trivy.go
```diff
@@ -49,23 +44,34 @@ type Results struct {
...
+	Class           string `json:"Class"`
+	Type            string `json:"Type"`
 	Vulnerabilities []struct {
 		VulnerabilityID  string   `json:"VulnerabilityID"`
 		PkgName          string   `json:"PkgName"`
+		PkgPath          string   `json:"PkgPath"`
 		InstalledVersion string   `json:"InstalledVersion"`
 		FixedVersion     string   `json:"FixedVersion"`
 		Title            string   `json:"Title,omitempty"`
 		Description      string   `json:"Description,omitempty"`
```
jesusfcr, line 56: We are not using Title / Description. It adds detailed information but it could be too much info.

```json
{
  "Title": "zlib: A flaw in zlib-1.2.11 when compressing (not decompressing!) certain inputs.",
  "Description": "zlib before 1.2.12 allows memory corruption when deflating (i.e., when compressing) if the input has many distant matches."}
```

cmd/vulcan-trivy/vulcan-trivy.go
danfaizer: I think is fair as long we are using the tool that provides the info.

cmd/vulcan-trivy/vulcan-trivy.go
danfaizer: The fingerprint for this vulnerability should be an array formed by:
The list of CVEs (not CWEs) + the list of severities of each CVE.
All sorted.

cmd/vulcan-trivy/vulcan-trivy.go
```diff
@@ -22,18 +22,13 @@ import (
...
 	checkstate "github.com/adevinta/vulcan-check-sdk/state"
 	"github.com/avast/retry-go"
-	"github.com/mcuadros/go-version"
 )
 
-const (
-	vulnTruncateLimit   = 30
-	vulnCVETrucateLimit = 10
-)
+const vulnCVETrucateLimit = 10
```
danfaizer, line 27: With the expansion of vulnerabilities now is true that we don't need to limit the number of vulnerabilities to 30. Because of this I would suggest to increase the number of `vulnCVETrucateLimit` to 30 instead.
As extra context, we would take the list of CVEs in order to compile the fingerprint of the vulnerability and as less information we loose the better.

cmd/vulcan-trivy/vulcan-trivy.go
danfaizer: In this case, in order to make the logic simpler and to make it more readable to the user I would suggest something like:
```
switch key.path {
 case "":
   vuln.AffectedResource = strings.TrimSpace(fmt.Sprintf("%s:%s:%s", tt.Type, key.name, key.version))
 default:
   vuln.AffectedResource = strings.TrimSpace(fmt.Sprintf("%s", key.path))
}
```
In this way, if there is a specific path where the vulnerability has been found (the detection point) it will be shown directly otherwise we will show required info to point to the correct place. From your examples:
```
# alpine:libretls:3.3.4-r2
# Java-flyway/lib/aad/jackson-databind-2.10.5.1.jar
# app/vulcan-api:github.com/docker/distribution:v2.7.1+incompatible
```

danfaizer (changes requested):
Great job! I've added some comments.

cmd/vulcan-trivy/vulcan-trivy.go
jesusfcr: Ok, but we could keep the list in the table to 10 and use the full list of CVEs for the finterprint.

cmd/vulcan-trivy/vulcan-trivy.go
```diff
@@ -186,116 +187,102 @@ func run(ctx context.Context, target, assetType, optJSON string, state checkstat
...
+			l, ok := outdatedPackageVulns[key]
+			if !ok {
+				l = []outdatedPackage{}
+			}
+			outdatedPackageVulns[key] = append(l, outdatedPackage{
+				severity: tv.Severity,
+				fixedBy:  tv.FixedVersion,
+				cve:      tv.VulnerabilityID,
+				link:     tv.PrimaryURL,
+				cves:     tv.CweIDs,
```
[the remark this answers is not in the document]
jesusfcr: Right, thanks!

cmd/vulcan-trivy/vulcan-trivy.go
```diff
@@ -186,116 +187,102 @@ func run(ctx context.Context, target, assetType, optJSON string, state checkstat
...
-					version:  dockerVuln.InstalledVersion,
-					severity: dockerVuln.Severity,
-					fixedBy:  dockerVuln.FixedVersion,
-					cves:     []string{},
+				row := make(map[string]string, len(vp.Header))
+				row["FixedBy"] = l.fixedBy
+				if l.cves != nil {
+					row["CWEs"] = strings.Join(l.cves, ",")
+				}
+				row["CVE"] = l.cve
```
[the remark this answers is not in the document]
jesusfcr: Done

cmd/vulcan-trivy/vulcan-trivy.go
jesusfcr: Sure, below we are hardcoding CWEID:937, but trivy provides CweIDs as an array.
I m wondering if we should compute the final CWEID from the ones provided by Trivy or keep the 937

cmd/vulcan-trivy/vulcan-trivy.go
```diff
@@ -186,116 +187,102 @@ func run(ctx context.Context, target, assetType, optJSON string, state checkstat
...
-		for _, dockerVuln := range trivyTarget.Vulnerabilities {
-			op, ok := outdatedPackageVulns[dockerVuln.PkgName]
-			if ok {
-				if isMoreSevere(dockerVuln.Severity, op.severity) {
-					op.severity = dockerVuln.Severity
+		vp := report.ResourcesGroup{
+			Name: "Package Vulnerabilities",
+			Header: []string{
+				"FixedBy",
+				"CVE",
```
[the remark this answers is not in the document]
jesusfcr: Done.

cmd/vulcan-trivy/vulcan-trivy.go
jesusfcr: I see two problems:

In case of jars packed the path exists but we willl miss the embedded package name / version"
```json
{
  "Target": "Java", "Class": "lang-pkgs", "Type": "jar",
  "Vulnerabilities": [
    {
      "VulnerabilityID": "CVE-2020-36518",
      "PkgName": "com.fasterxml.jackson.core:jackson-databind",
      "PkgPath": "app/app.jar",
      "InstalledVersion": "2.13.0",
      "FixedVersion": "2.12.6.1, 2.13.2.1"
    }
}
```

In case of gobinaries the path is in the tt.Target
```json
{
  "Target": "app/vulcan-api", "Class": "lang-pkgs", "Type": "gobinary",
  "Vulnerabilities": [
    {
      "VulnerabilityID": "GMS-2022-20",
      "PkgName": "github.com/docker/distribution",
      "InstalledVersion": "v2.7.1+incompatible",
      "FixedVersion": "v2.8.0"
    }
}
```

jesusfcr: I propossed a variation.

cmd/vulcan-trivy/vulcan-trivy.go
danfaizer: I think is ok to have these attributes mapped at check level even if we not use them for now.
We just need to pay attention if we decide to use them in the future in terms of vulnerability size.

cmd/vulcan-trivy/vulcan-trivy.go
jesusfcr: Kept the 10 limit and using the whole list for the fingerprint.

gguillemas (dismissed):
LGTM

danfaizer (dismissed):
LGTM

danfaizer (approved):

PR #179, opened 2022-03-28
Update vulcan-lucky to v2

octomit:
Update vulcan-lucky to v2

### DIFF
--- a/cmd/vulcan-lucky/main.go
+++ b/cmd/vulcan-lucky/main.go
@@ -1,5 +1,5 @@
 /*
-Copyright 2022 Adevinta
+Copyright 2019 Adevinta
 */
 
 package main
@@ -32,7 +32,7 @@ var (
 		AffectedResource: "443/tcp",
 		Score:            report.SeverityThresholdMedium,
 		Fingerprint:      helpers.ComputeFingerprint(),
-		Labels:           []string{"issue"},
+		Labels:           []string{"issue", "ssl"},
 		ImpactDetails:    "Allows remote attackers to obtain sensitive cleartext information via a padding-oracle attack against an AES CBC session",
 		References:       []string{"https://blog.cloudflare.com/yet-another-padding-oracle-in-openssl-cbc-ciphersuites/"},
 		Recommendations:  []string{"Upgrade OpenSSL to, at least, 1.0.2h or 1.0.1t"},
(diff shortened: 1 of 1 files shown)


cmd/vulcan-lucky/main.go
```diff
@@ -1,5 +1,5 @@
 /*
-Copyright 2019 Adevinta
+Copyright 2022 Adevinta
```
julianvilas, line 2: I think this is not required. We would put the 2022 header just if the file has been created on 2022. But I might be wrong.

cmd/vulcan-lucky/main.go
```diff
@@ -29,10 +29,13 @@ var (
...
 		Description: "AES-NI implementation in OpenSSL before 1.0.1t and 1.0.2 before 1.0.2h allows remote attackers to obtain sensitive" +
 			"cleartext information",
-		Score:           report.SeverityThresholdMedium,
-		ImpactDetails:   "Allows remote attackers to obtain sensitive cleartext information via a padding-oracle attack against an AES CBC session",
-		References:      []string{"https://blog.cloudflare.com/yet-another-padding-oracle-in-openssl-cbc-ciphersuites/"},
-		Recommendations: []string{"Upgrade OpenSSL to, at least, 1.0.2h or 1.0.1t"},
+		AffectedResource: "443/tcp",
+		Score:            report.SeverityThresholdMedium,
+		Fingerprint:      helpers.ComputeFingerprint(),
+		Labels:           []string{"issue"},
```
julianvilas, line 35: I'd suggest also adding the optional `ssl` label, as we have added it in some other checks reporting problems related to SSL.

julianvilas (commented):
Very good job man! thx for the contribution. Just a couple of minor suggestions

cmd/vulcan-lucky/main.go
octomit: Addressed in 83d2167c389d0f4be6f076ccd271a29ad32e1c77

cmd/vulcan-lucky/main.go
octomit: Done. 
83d2167c389d0f4be6f076ccd271a29ad32e1c77

julianvilas (approved):

PR #180, opened 2022-03-28
Update vulcan-drupal to v2

octomit:


cmd/vulcan-drupal/vulnerabilities.go
```diff
@@ -22,21 +25,27 @@ var (
...
 			Vulnerability: report.Vulnerability{
-				Summary:         "Drupal - End-of-Life",
-				CWEID:           937,
-				Description:     "Versions of Drupal 9 prior to 9.1, 8 prior to 8.9 and versions prior to 7 are end-of-life and do not receive security coverage.",
-				Score:           report.SeverityThresholdCritical,
-				Recommendations: []string{"Update to a supported version"},
+				Summary:          "Drupal - End-of-Life",
+				CWEID:            937,
+				AffectedResource: "443/tcp",
+				Fingerprint:      helpers.ComputeFingerprint(),
```
julianvilas, line 31: If you check [this part of the code](https://github.com/adevinta/vulcan-checks/blob/fceaa151dc199311da63dae22867c91e2d83d5bd/cmd/vulcan-drupal/main.go#L134), you'll realize that the issue might be found also in the port `80/tcp`, so you need to calculate the `AffectedResource` in runtime.

So for example in the case of the [infoDrupal vulnerability](https://github.com/adevinta/vulcan-checks/pull/180/files#diff-b31fbe7c0864dcd4d3b657daa0c1c47ccc0145880bdf898e83d6813584e6c5aeR18) (where you missed to add the labels too) you might want to add this logic in [main.go#L156](https://github.com/adevinta/vulcan-checks/blob/fceaa151dc199311da63dae22867c91e2d83d5bd/cmd/vulcan-drupal/main.go#L156):
```
infoDrupal.AffectedResource = u.String()
```
Also I'd compute in runtime the fingerprint, using the detected version:
```
infoDrupal.Fingerprint = helpers.ComputeFingerprint(version)
```

The same recommendation applies to all the other vulnerabilities, but for those other cases the change would in [main.go#L48](https://github.com/adevinta/vulcan-checks/blob/fceaa151dc199311da63dae22867c91e2d83d5bd/cmd/vulcan-drupal/main.go#L48) before adding it to the slice. So something like:
```
v.Vulnerability.AffectedResource = u.String // You need to pass this to the method as currently is not passed.
v.Vulnerability.Fingerprint = helpers.ComputeFingerprint(ver) // Alternatively you might calculate the fingerprint value in a temp var just before the loop and assign it here to avoid doing it more than once.
vulnerabilities = append(vulnerabilities, v.Vulnerability)
```

cmd/vulcan-drupal/vulnerabilities.go
```diff
@@ -87,14 +102,17 @@ var (
...
+				Labels:           []string{"issue"},
+				Description:      "The QuickEdit module does not properly validate access to routes, which could allow cross-site request forgery under some circumstances and lead to possible data integrity issues. Sites are only affected if the QuickEdit module (which comes with the Standard profile) is installed. Removing the \"access in-place editing\" permission from untrusted users will not fully mitigate the vulnerability. This advisory is not covered by Drupal Steward.",
+				Score:            report.SeverityThresholdMedium,
+				References:       []string{"https://www.drupal.org/sa-core-2021-007"},
 				Recommendations: []string{
 					"Install the latest version:" +
-					"If you are using Drupal 9.2, update to Drupal 9.2.6.",
+						"If you are using Drupal 9.2, update to Drupal 9.2.6.",
```
julianvilas, line 115: Is it me or this got an extra-indentation?

julianvilas (changes requested):
Hey man, thanks for the contribution, great work! I put some comments to address.

julianvilas (approved):

PR #182, opened 2022-04-04
Improve WPScan execution behaviour

danfaizer:
This PR tries to improve several aspects of the execution of WPScan command.
- Specified a custom HTTP user-agent.
- Add `--ignore-main-redirect` param if the target redirects to a site contains any matching pattern specified in `ignoreMainRedirect`.
- In the case WPScan is aborted due to WAF user-agent header detection, we are trying again with `--random-user-agent` parameter as is recommended by the tool.
- WPScan cli tool exit code is now being taken into count for the result.

### DIFF
--- a/cmd/vulcan-wpscan/main.go
+++ b/cmd/vulcan-wpscan/main.go
@@ -55,6 +55,10 @@ func main() {
 		}
 
 		wpScanReport, err := RunWpScan(ctx, logger, target, url, token)
+		if err != nil && err.Error() == NotAWordPressMessage {
+			logger.Warnf("%s", err)
+			return checkstate.ErrAssetUnreachable
+		}
 		if err != nil {
 			return err
 		}
--- a/cmd/vulcan-wpscan/main.go
+++ b/cmd/vulcan-wpscan/main.go
@@ -55,9 +55,10 @@ func main() {
 		}
 
 		wpScanReport, err := RunWpScan(ctx, logger, target, url, token)
+		// If the target is not a WordPress site finish the check gracefully.
 		if err != nil && err.Error() == NotAWordPressMessage {
-			logger.Warnf("%s", err)
-			return checkstate.ErrAssetUnreachable
+			logger.Infof("%s", err)
+			return nil
 		}
 		if err != nil {
 			return err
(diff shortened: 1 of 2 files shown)


cmd/vulcan-wpscan/main.go
```diff
@@ -55,6 +55,10 @@ func main() {
 		}
 
 		wpScanReport, err := RunWpScan(ctx, logger, target, url, token)
+		if err != nil && err.Error() == NotAWordPressMessage {
+			logger.Warnf("%s", err)
+			return checkstate.ErrAssetUnreachable
+		}
```
julianvilas, line 61: I think we are misusing the `INCONCLUSIVE` status here. The intent of that status is to detect stale assets so we might notify the end user that some action is required by them (e.g. update or remove). It would be appropriate if we had an asset type like `WordPress` but we don't. Every check ends up `FINISHED` with empty `[]report.Vulnerability` when no issues are found.

cmd/vulcan-wpscan/wpscan.go
```diff
@@ -26,11 +33,18 @@ var (
 	wpscanForceParams = []string{"--force", "--wp-content-dir", "wp-content"}
 	wpscanScopeParams = []string{"--scope"}
 	wpscanTokenParams = []string{"--api-token"}
+	wpscanUserAgent   = []string{"--user-agent", "Vulcan"}
+
+	// --ignore-main-redirect string matching list.
```
julianvilas, line 38: I think it would be good to explain a bit the rationale for this change so in the future we know why is this here.

julianvilas (changes requested):
Great job! :) 

Just requesting a change and suggesting to document a bit the rationale for the new features.

cmd/vulcan-wpscan/main.go
danfaizer: Right, I see. Suggestion applied here [20267b9](https://github.com/adevinta/vulcan-checks/pull/182/commits/20267b9bd2809f4e13d38857ecd732a4ac46791c)

cmd/vulcan-wpscan/wpscan.go
danfaizer: Yup, good suggestion. Explanation added here [5e7a74d](https://github.com/adevinta/vulcan-checks/pull/182/commits/5e7a74d58e579d8cfcb685df7a5a480f3115bad4)

julianvilas (approved):
LGTM!

PR #185, opened 2022-04-06
Add gitleaks

mmmgili:


cmd/vulcan-gitleaks/gitleaks.go
```diff
@@ -0,0 +1,59 @@
...
+import (
+	"context"
+
+	"github.com/adevinta/vulcan-check-sdk/helpers/command"
+	"github.com/sirupsen/logrus"
+)
+
+const Cmd = `gitleaks`
+
+var params = []string{"detect", "/tmp/repo", "-f", "json", "-r", reportOutputFile, "--no-git"}
```
svalentino, line 16: One suggestion here, but please feel free to discard it. 

Instead of running the tool with the default and embedded rule set, why not passing a custom [.gitleaks.toml](https://github.com/zricethezav/gitleaks/blob/master/config/gitleaks.toml) file? The advantages I see are:
1. we can easily disable/customise wanky rules (and that's actually the case for [this one rule](https://github.com/zricethezav/gitleaks/blob/master/config/gitleaks.toml#L627-L640) that I personally keep disabled due to the sheer amount of false positive it raises. We can talk about this specific rule in a different thread if you want)
2. more importantly, we can easily add our own custom rules to the stack. I actually have a working example I've had to use for the pentest of Fotocasa.pro:

   ```
    [[rules]]
    id = "sql-connection-string"
    description = "Connection string for a SQL database"
    regex = '''(?m)connectionString=".*password=([^;]+)'''
    secretGroup = 1
   ```
    (yes, I know - I should submit a PR upstream for this, but I am lazy 😅)

The only disadvantage I see for this approach is that getting the latest updates from the upstream version of `.gitleaks.toml` file becomes then a manual job.

@everyone - thoughts?

cmd/vulcan-gitleaks/main.go
```diff
@@ -0,0 +1,237 @@
...
+	r, err := git.PlainClone(repoPath, false, &co)
+	if err != nil {
+		return err
+	}
+
+	b, err := r.Head()
+	if err != nil {
+		logger.Errorf("vulcan-gitleaks was not able to retrieve the working branch")
+		return err
+	}
```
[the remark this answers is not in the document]
gguillemas: Marc mentioned this very same thing to me a couple of days ago and I completely agree 👍

cmd/vulcan-gitleaks/Dockerfile
```diff
@@ -0,0 +1,12 @@
+# Copyright 2019 Adevinta
+
+FROM zricethezav/gitleaks
```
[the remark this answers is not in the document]
gguillemas: I agree with your concern, I think any of your proposed solutions are good!

cmd/vulcan-gitleaks/gitleaks.go
gguillemas: I don't have a strong opinion on this and I am personally okay with using the embedded rule set. However, we have an example of a check that functions exactly how you describe: https://github.com/adevinta/vulcan-checks/blob/master/cmd/vulcan-exposed-http-resources/main.go#L135-L145
In any case, I think accepting rules to be supplied through the options can always be useful.

cmd/vulcan-gitleaks/main.go
```diff
@@ -0,0 +1,237 @@
...
+		Labels:        []string{"issue"},
+		Recommendations: []string{
+			"Completely remove the secrets from the repository as explained in the references.",
+			"Encrypt the secrets using a tool like AWS Secrets Manager or Vault.",
+		},
+		References: []string{
+			"https://help.github.com/en/articles/removing-sensitive-data-from-a-repository",
+		},
+	}
+	details  = "This secret was found by the gitleaks rule '%s', with ID '%s'. If this doesn't correspond to a secret, it can be excluded from future scans marking it as false positive, or adding the following line to vulcan.yaml file for vulcan-local executions:\n \t- affectedResource: %s"
```
julianvilas, line 54: I'm unsure about mentioning the `vulcan.yaml` here, as it isn't part of the check itself and it isn't used in regular scans. I think that kind of information should be provided by vulcan-local itself.

cmd/vulcan-gitleaks/manifest.toml
```diff
@@ -0,0 +1,5 @@
+Description = "Finds potential secrets in source code from a Git repository using gitleaks"
+Timeout = 900 # 15 minutes
+AssetTypes = ["GitRepository"]
+RequiredVars = []
```
julianvilas, line 4: I think we are missing the required vars to be able to clone the repo:
```RequiredVars = ["GITHUB_ENTERPRISE_ENDPOINT", "GITHUB_ENTERPRISE_TOKEN"]```

File: cmd/vulcan-gitleaks/main.go [hunk 6]
```diff
@@ -0,0 +1,237 @@
...
+	for _, f := range results {
+		if stringInSlice(f.RuleID, &opt.ExcludedRules) {
+			continue
+		}
+		file := strings.TrimPrefix(f.File, repoPath)
+		v := leakedSecret
+		s, _ := bcrypt.GenerateFromPassword([]byte(f.Secret), 0)
+		v.AffectedResource = string(s)
+		v.AffectedResourceString = computeAffectedResource(target, branch, file, f.StartLine)
+		v.Fingerprint = helpers.ComputeFingerprint("")
```
julianvilas, line 202: Not relevant at all, but I'd suggest removing the `""` and just call the method without args, to be consistent with what has been implemented in the other checks.

cmd/vulcan-gitleaks/main.go
```diff
@@ -0,0 +1,237 @@
...
+	return processVulns(results, opt, repoPath, string(b.Name()), target, state)
+}
+
+func processVulns(results []Finding, opt options, repoPath string, branch string, target string, state checkstate.State) error {
+	// Return if there are no findings.
+	if len(results) < 1 {
+		return nil
+	}
+
+	// Group secrets by file.
```
julianvilas, line 192: This comment is not true anymore, right?

julianvilas (changes requested):
Thanks for the great contribution! :)

Just adding some more comments.

cmd/vulcan-gitleaks/Dockerfile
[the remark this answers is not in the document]
julianvilas: This is a very interesting topic that deserves a discussion. We are using the same approach currently used in this check in some other checks and probably in some other places different from Vulcan checks.

I think the problem in this case is that we are using the default rules they provide, and [that file]((https://github.com/zricethezav/gitleaks/blob/master/config/gitleaks.toml)) is included in the docker image itself. As discussed in another comment below, we might not use their rules, but then we need to be on top of that (operational work) and this is not working very well for us.

So another option would be to pin a version and then get the default ruleset file dynamically or during build time.

Do you think, given that this is not the only place where this is happening, we might merge with this current approach and discuss about this in order to find a solution that works in general?

cmd/vulcan-gitleaks/main.go
gguillemas: I think that is a good point. I would personally keep the `vulcan.yaml` reference only if we integrated it somehow in the check itself. Otherwise this information only applies to `vulcan-local`. However, if we plan on running the check through the regular Vulcan scans I feel we should allow users to whitelist secrets proactively from a file (as with the Seekret check) and not only from the "Mark as False Positive" feature. I think it may make sense to add some ability to parse `vulcan.yaml` in the checks SDK and use that to filter those findings directly in the check. This has some implications and probably is worth it to discuss separately.

cmd/vulcan-gitleaks/main.go
mmmgili: That's true, removed here [ec07e44](https://github.com/adevinta/vulcan-checks/pull/185/commits/ec07e443a7a05c55ec762fc6ffe12642f82d8256).

cmd/vulcan-gitleaks/main.go
mmmgili: Completelly agree, fixed here [ec07e44](https://github.com/adevinta/vulcan-checks/pull/185/commits/ec07e443a7a05c55ec762fc6ffe12642f82d8256).

cmd/vulcan-gitleaks/manifest.toml
mmmgili: Yes, I added them here [ec07e44](https://github.com/adevinta/vulcan-checks/pull/185/commits/ec07e443a7a05c55ec762fc6ffe12642f82d8256).

cmd/vulcan-gitleaks/main.go
mmmgili: I agree with you, this detail will be added only when scanning local resources (done here [ec07e44](https://github.com/adevinta/vulcan-checks/pull/185/commits/ec07e443a7a05c55ec762fc6ffe12642f82d8256)), since the only way of doing so for now is vulcan-local. Do you think this change is enough or would you remove the references in all of the cases?

cmd/vulcan-gitleaks/gitleaks.go
mmmgili: Another option would be to add a custom `.gitleaks.toml` file to the workdir of the Docker image, and the tool will automatically use this file as the ruleset. This would require us to review changes in the upstream configuration file to update our custom file with latest rules.
Allowing additional rules to be provided through options can be tricky but it's possible, although I don't see a clear advantage of doing so in front of creating a custom ruleset file, taking into account the complexity this can add to the check manteinance.
Regarding the rule disabling, it is currently supported through the ExcludedRules option.

cmd/vulcan-gitleaks/main.go
gguillemas: In my opinion this is enough. I don't know our stance in referencing Vulcan API features such as false positives on the checks, but I personally think it is fine and helpful in the current use cases. Great job 👌

cmd/vulcan-gitleaks/gitleaks.go
svalentino: Related to [Julian's comment](https://github.com/adevinta/vulcan-checks/pull/185#discussion_r847348150), I think it's fine to use the predefined rule list for now (I mean, this is purely a product decision), this way we also avoid the toil of manually merging changes from upstream (which, as Julian mentioned, it doesn't really work for us). 

This commend can be resolved I think.

cmd/vulcan-gitleaks/Dockerfile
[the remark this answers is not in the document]
mmmgili: In this case, as a intermediate solution, we could pin the gitleaks Docker image version [configure Dependabot](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#docker-registry) to check if there's a new version of the image available.
The problem I see with this solution is that we might be spammed with PRs, since [gitleaks has been creating a new release](https://github.com/zricethezav/gitleaks/releases) almost every week.

julianvilas (approved):

PR #186, opened 2022-04-11
bugfix: remove informational label from all checks

julianvilas:
The fact that the same vulnerability might have different labels per execution was introducing some unexpected behaviours, as the labels are defined at _issue_ level in the Vulnerability DB model. This was particularly triggered in those cases where the same issue might have a score 0 or different from 0 depending on the execution.

It was discussed an agreed to not label vulnerabilities as `informational` when the score is 0. We will label them with the exact same labels than when the score is different from 0. Even more, labels shouldn't change during runtime, as they shouldn't include runtime information.


danfaizer (approved):
LGTM

PR #187, opened 2022-05-06
Support configure ignored scanners for fingerprint generation

ka3de:
This PR adds a new supported option for `vulcan-zap` check which allows to define a list of Zap proxy plugin IDs for which the resources associated with the vulnerabilities reported by them will not be taken into account for the fingerprint generation.

The reason for this functionality is to try to mitigate the variability of the fingerprint for some type of vulnerabilities, especially the injection type ones like SQLi, which eventually does not allow Vulcan end users to mark a finding as _false positive_ permanently, as the finding is recurrently being reopened due to the variations on the associated resources, for example the payload used for the attack.


danfaizer (approved):
Smooth! 👍 
Thanks for the contribution @ka3de 🙇

PR #188, opened 2022-05-25
Force status 200 for Laravel

jesusfcr:
When Laravel is in debug mode it generates a 404 page showing the source code of the exception

```php
$kernel = new AppKernel($_SERVER['APP_ENV'], (bool) $_SERVER['APP_DEBUG']);
$request = Request::createFromGlobals();
$response = $kernel->handle($request);$response->send();
```

Another option could be to remove APP_ENV from the regex


danfaizer (approved):
LGTM

PR #189, opened 2022-05-26
Deprecate vulcan-dkim check

### LINKED ISSUE #72: false positive for dkim check on schibsted.com
The dkim check reports a false positive on schibsted.com, due to that the dkim TXT record is on the dns field `google._domainkey.schibsted.com` instead of `default._domainkey.schibsted.com`.

The dns name can of course be anything, but it would be good to add a check for `google._domainkey` also.

I read through the code, and it might just be to add another string here?

https://github.com/adevinta/vulcan-checks/blob/master/cmd/vulcan-dkim/local.toml.example#L5

danfaizer:
This PR is meant to resolve #72 

After doing a review about the current implementation of the check and how DKIM is implemented by different providers we've concluded that is not possible to have a check that adds value reporting missing DKIM configuration.
We could use have used `google._domainkey.$domain` as @alexanderkjall proposed in order to check if DKIM is configured for Google mail but still, this is not the only and valid indicator DKIM is well configured for other mail providers and therefore there is no much real value (more than informational) in reporting that.

Because of this we've decided to deprecate this check.


gguillemas (approved):
LGTM

PR #191, opened 2022-06-01
upgrade

jesusfcr:
- Upgrade to 1.18 and tidy
- Pin adevinta deps


go.mod
```diff
@@ -1,45 +1,64 @@
...
 	github.com/FiloSottile/CVE-2016-2107 v0.0.0-20160513191821-17e9b6082f41
 	github.com/FiloSottile/Heartbleed v0.2.1-0.20150408030656-4a3332ca1dc0
 	github.com/Masterminds/semver/v3 v3.1.0
-	github.com/Microsoft/go-winio v0.5.1 // indirect
-	github.com/adevinta/gozuul v0.0.0-20191210100135-8808882dea1f
-	github.com/adevinta/restuss v0.0.0-20200401171945-cc64e2b9dd21
-	github.com/adevinta/vulcan-check-sdk v0.0.0-20220607152912-cca827244598
-	github.com/adevinta/vulcan-report v0.0.0-20211117082128-cadc974cc14c
-	github.com/adevinta/vulcan-types v0.0.0-20220310133120-53f000e5196f
+	github.com/adevinta/gozuul v0.0.0-20210319091150-d32db4978cf4
```
ka3de, line 9: Why is gozuul not using semver?

jesusfcr: Yes, that's correct, we discussed with @danfaizer and we agreed it doesn't worth the effort.

PR #199, opened 2022-08-04
Remove X-XSS-Protection findings

kozmic:
As X-XSS-Protection header is deprecated and no longer recommended, I think it should be removed from this check. This change only make it ignore X-XSS-Protection findings, and does not change [the struct](https://github.com/adevinta/vulcan-checks/blob/master/cmd/vulcan-http-headers/observatory.go#L126-L133) `observatoryResult`.


kozmic:
There is an open issue upstream in http-observatory: https://github.com/mozilla/http-observatory/issues/432

gguillemas (approved):
Thanks for the PR! 👍

PR #203, opened 2022-09-21
solve severity and hash problem in gitleaks check

eloymg:
Change severity to 8.9 to be a real HIGH and solve the problem that bcrypt algorithm generates a different hash in every execution because the salt is random generated.


jesusfcr (approved):
LGTM

PR #211, opened 2022-10-10
Solve problem with Cwe in semgrep check

eloymg:
Comment Cwe field for the problem that can be and array of strings or string, and this cause that the Check Fail i some cases.


jesusfcr (commented):
Thanks for your pr
The fix does't works because Cwe is referenced.
I made this PR #213

eloymg:
@jesusfcr Thanks! I close this PR.

PR #213, opened 2022-10-11
fix semgrep

jesusfcr:
- Fix Unmarshal adding support for slices in cwe
- Pin semgrep 0.116.0
- Use helpers.CloneGitRepository in semgrep

### DIFF
--- a/cmd/vulcan-semgrep/main.go
+++ b/cmd/vulcan-semgrep/main.go
@@ -249,7 +249,17 @@ func vuln(result Result, filepath string, vulns map[string]report.Vulnerability)
 	summary := issue
 
 	var cweID int
-	if matches := CWERegex.FindStringSubmatch(result.Extra.Metadata.Cwe); len(matches) == 3 {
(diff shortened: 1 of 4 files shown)


cmd/vulcan-semgrep/main.go
```diff
@@ -259,7 +215,7 @@ func vuln(result Result, filepath string, vulns map[string]report.Vulnerability)
 		cweID, _ = strconv.Atoi(matches[1])
 	}
 
-	summary = strings.Title(summary)
+	summary = cases.Title(language.English).String(summary)
```
julianvilas, line 218: Does this change any of the existent summaries? Take into consideration that a change on the summary is a change on the issue and therefore we loose track of the findings.

jesusfcr: Good catch! It shouldn't but not 100% sure. I reverted the change.

julianvilas (approved):

PR #219, opened 2022-11-04
Fix incorrect use of references and quote encoding

gguillemas:
This PR resolves a couple of issues with the exposed files check:

- The references are issue-level and shown for all findings so should not contain any custom data.
- The quotes were double escaped and the backslashes were displayed in the UI.

This check may be further improved by merging all vulnerabilities into one with several resources listed in the resources table as other checks do. However, I have chosen not to do that since this check may soon either be deprecated (in favor of Nuclei) or merged with the exposed HTTP resources check which is very similar.


ka3de (approved):
LGTM.

PR #222, opened 2022-11-14
Fix tenable for multiplatform

jesusfcr:


danfaizer (approved):
Thanks for the fix! :bow:

PR #225, opened 2022-11-15
Use container openssl available in all architectures

jesusfcr:
The check relies on `openssl` binaries from https://github.com/mozilla/cipherscan, old version and only available for Darwin or amd64 (not arm64)
This pr forces the use of the openssl version in `python:3.7-slim`

Upgrades the pretty old openssl version
```sh
# ./openssl version
OpenSSL 1.0.2-chacha (1.0.2i-dev)
# openssl version
OpenSSL 1.1.1n  15 Mar 2022
```

### DIFF
--- a/cmd/vulcan-tls/main.go
+++ b/cmd/vulcan-tls/main.go
@@ -53,10 +53,6 @@ type failure struct {
 	Issues []string
 }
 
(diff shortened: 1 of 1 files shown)


danfaizer (approved):
:heavy_check_mark:

PR #230, opened 2022-12-13
Fix Nuclei flag error and set conservative templates

gguillemas:
This PR fixes an error with a [deprecated Nuclei flag](https://github.com/projectdiscovery/nuclei/pull/2962), pins the current version to avoid future breakages and enables a limited list of conservative templates so that the check can be included in the default Vulcan policy. Further templates can be included in more aggressive policies or for specific assets by adding them through options or removing the whitelist option altogether.


mmmgili (approved):
LGTM!

PR #232, opened 2022-12-19
Added Njsscan check

eloymg:
I created a check for execute [Njsscan Check](https://github.com/ajinabraham/njsscan)


jesusfcr:
`njsscan` is based on semgrep rules `p/nodejsscan`.

After your last contribution to the semgrep check we can get the same results with semgrep.

```
vulcan-local -t . -i semgrep \
  -o '{"ruleset":["r/typescript","r/javascript","r/nodejsscan","r/r2c-security-audit","p/nodejsscan"]}'
```

So I think there is no reason to create a specific check for that purpose.

eloymg:
For the reasons that mention jesusfcr, I close the PR

PR #233, opened 2022-12-21
Correctly parse Github Alerts GraphQL output

gguillemas:
Vulcan is reporting previously reported vulnerabilities detected by Github as open when in fact they have been fixed or dismissed. This happens because the Github GraphQL API returns all existing and previous vulnerabilities regardless of current status. This issue is fixed by explicitly verifying that the current status of the vulnerability is "OPEN". This requires pulling the additional "[RepositoryVulnerabilityAlertState](https://docs.github.com/en/graphql/reference/enums#repositoryvulnerabilityalertstate)" attribute from GraphQL and checking it reporting a vulnerability.

Additionally, the "[FirstPatchedVersion](https://docs.github.com/es/graphql/reference/objects#securityadvisorypackageversion)" attribute was being parsed as a string instead of an struct due to it being missing from all responses returned by Github. Now that it is being returned, some checks are failing due to incorrect parsing. This is resolved by parsing it correctly and relying on the value provided if it is present in the response.

### DIFF
--- a/cmd/vulcan-github-alerts/main.go
+++ b/cmd/vulcan-github-alerts/main.go
@@ -71,29 +71,26 @@ type alertsData struct {
 type Details struct {
 	State                 string `json:"state"`
 	SecurityVulnerability struct {
-		Advisory ExtendedAdvisory `json:"advisory"`
+		Advisory Advisory `json:"advisory"`
 		Package  struct {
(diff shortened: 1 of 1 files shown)


danfaizer (approved):
Awesome! Thanks!

PR #240, opened 2023-01-18
Adapt Nuclei check to support custom templates

gguillemas:
Adapt Nuclei check to support custom templates by doing the following:

- Update Nuclei to the latest version.
- Add verbosity to Nuclei at the build stage to catch issues.
- Add templates in the `templates` directory to the [Nuclei default templates](https://github.com/projectdiscovery/nuclei-templates/). Overwrite conflicting ones.
- Create a `tests` directory that will contain scripts to test custom templates.
- Include several PRs replacing old checks for default and custom Nuclei templates.


cmd/vulcan-nuclei/main.go
```diff
@@ -134,16 +134,10 @@ func processNucleiFindings(target string, nucleiFindings []ResultEvent) []*repor
 	vulnerabilities := []*report.Vulnerability{}
 	rv := make(map[string]*report.Vulnerability)
 	for _, v := range nucleiFindings {
-		// Create resources table row.
-		// Avoid store redundant information in the resources table.
-		matched := v.Matched
-		if matched == target {
-			matched = ""
-		}
```
julianvilas, line 142: What's the reason to remove this code?

cmd/vulcan-nuclei/main.go
```diff
@@ -134,16 +134,10 @@ func processNucleiFindings(target string, nucleiFindings []ResultEvent) []*repor
...
 		findingRow := map[string]string{
 			"Template":    v.TemplateID,
 			"MatcherName": v.MatcherName,
-			"Matched":     matched,
+			"Matched":     v.Matched,
```
julianvilas, line 140: What's the reason to change this?

cmd/vulcan-nuclei/main.go
```diff
@@ -167,7 +161,7 @@ func processNucleiFindings(target string, nucleiFindings []ResultEvent) []*repor
 			}
 		}
 		var vuln = report.Vulnerability{
-			AffectedResource: target,
+			AffectedResource: v.Matched,
```
julianvilas, line 164: I'd really like to see if there's any kind of variability of this field between executions.

cmd/vulcan-nuclei/main.go
gguillemas: The reason is that after the change below the `matched` variable was not being used. I will explain why I decided in favor of this change in the other comment you wrote.

cmd/vulcan-nuclei/main.go
gguillemas: The reason is that `matched` was empty when the target string was the same as the matched string. This never would happen with hostname assets because the matched string contains the port and the target does not. However, this would happen with web address assets, for instance in the case where the target is `https://example.com/` and the issue is found in the `/` path. In this case, I see no reason to leave the column empty whereas all the others will contain the full URL where the issue was found (e.g. `https://example.com/vulnerable/`). If we do that, we are leading the users to believe that the table is incomplete instead of clearly telling them in which path the issue was found.

cmd/vulcan-nuclei/main.go
gguillemas: We can do some tests on this, but it should not vary at all unless there is an actual change in the target. For network checks (such as the ones we are currently discussing in #241, #242, #243) this string will just be the hostname plus the port. In the case of web addresses the full URL where the check found the issue. It could happened that, for web applications, the vulnerability can be found in different places and different ones are reported each time, similar to what happens with the ZAP and Burp checks where the crawling is not exhaustive nor deterministic. However, this still means that the vulnerability was found in different places.

cmd/vulcan-nuclei/main.go
julianvilas: Gotcha, very good point.

cmd/vulcan-nuclei/main.go
julianvilas: Cool, thx.

julianvilas (approved):

gguillemas:
Commits up to this point have been reviewed and approved in the dismissed review and in their respective PRs.

julianvilas (approved):

PR #250, opened 2023-01-30
Fix vulcan prowler

danfaizer:
- Pin latest compatible version for the base docker image
- Add missing control
- Avoid fail the check if a control is missing


gguillemas (approved):
👍

PR #292, opened 2023-05-08
tests

jesusfcr:
This PR adds a mechanism to create tests for the check images.

- In case of an executable `test.sh` file exists in the check path it would be executed with the capabilities to break the process by exiting with error.
- The examples provided are based on vulcan-local but other tools can be used.
- In case of a failing test it stops
- The test is done in the deploy process, it will push images until a test fails.


release.sh
```diff
@@ -126,6 +131,25 @@ for check in "${CHECKS[@]}"; do
         "cmd/$check" --push
 
     log_msg "Builded image $check:[${IMAGE_TAGS[*]}]"
+
+    if [ -f "cmd/$check/vulcan.yaml" ]; then
+        log_msg "Testing image $check"
+        vulcan-local -i "$check" -c "cmd/$check/vulcan.yaml" -s CRITICAL -checktypes checktypes.json -l DEBUG -r - 2> error.log 1> r.json || true
+        STATUS=$(jq < r.json -r '.[].status' | tr '\n' ',')
+        if [[ ! $STATUS =~ ^(FINISHED|INCONCLUSIVE|,)+$ ]]; then
```
jesusfcr, line 139: Probably INCONCLUSIVE should't be a success state, but hard to generate a working test for some checks (i.e. prowler)

danfaizer (approved):
Very good job mate. Like it!

PR #381, opened 2023-07-21
Prevent vulcan-retirejs failing under some circumstances

julianvilas:
We have detected that `vulcan-retirejs` was failing under some circumstances:

1. While scanning sites with non-valid SSL certificates (expired, self-signed, etc.) (fixed in ffd74f6dc6e62c5dc721fd43e770fe8b9dd9012a)
2. When imported Javascript dependencies had several parameters (fixed in 37e082657ef6d822be1fcca6f68dbc224301e0fd)

This PR addresses those errors.

### DIFF
--- a/cmd/vulcan-retirejs/main.go
+++ b/cmd/vulcan-retirejs/main.go
@@ -376,13 +376,13 @@ func writeFile(fileName string, contents string) error {
 }
 
 func downloadFromUrl(URL string) error {
-	filename := ""
 	u, err := url.ParseRequestURI(URL)
-	if err == nil {
-		path := u.EscapedPath()
-		tokens := strings.Split(strings.Trim(path, "/"), "/")
-		filename = tokens[len(tokens)-1]
+	if err != nil {
+		return fmt.Errorf("error invalid url %s: %w", URL, err)
 	}
+	tokens := strings.Split(strings.Trim(u.EscapedPath(), "/"), "/")
+	filename := tokens[len(tokens)-1]
+
 	// NOTE: a random UUID is appended to avoid file path clashing with other
 	// previously downloaded scripts (e.g. '/a/jquery.min.js' and
 	// '/b/jquery.min.js').
--- a/cmd/vulcan-retirejs/main.go
+++ b/cmd/vulcan-retirejs/main.go
@@ -373,26 +373,27 @@ func writeFile(fileName string, contents string) error {
 	return nil
 }
 
-func downloadFromUrl(url string) error {
-	filePath := getFilePath(url)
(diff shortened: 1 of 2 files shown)


cmd/vulcan-retirejs/main.go
```diff
@@ -363,38 +375,31 @@ func writeFile(fileName string, contents string) error {
...
+	u, err := url.ParseRequestURI(URL)
+	if err == nil {
+		path := u.EscapedPath()
+		tokens := strings.Split(strings.Trim(path, "/"), "/")
+		filename = tokens[len(tokens)-1]
+	}
+	// NOTE: a random UUID is appended to avoid file path clashing with other
+	// previously downloaded scripts (e.g. '/a/jquery.min.js' and
+	// '/b/jquery.min.js').
+	filePath := fmt.Sprintf("%s/%s_%s.js", jsPath, filename, uuid.NewV4().String())
```
manelmontilla, line 389: It's not a big deal, but, If I'm getting this correctly, the `url` param can be an invalid URI, in which case, the function `url.ParseRequestURI` will return and error and cause this var `filePath` to get a value like: `temp/_UUID.js` which is a little bit odd. It is true that then call `response, err := client.Get(URL)` will return an error and that value will be discarded. But I'm curious why you didn't want to just do it in the more usual way:
```go
u, err := url.ParseRequestURI(URL)
	if err != nil {
		return fmt.Errorf("error invalid url %s: %v, URL, err)
	}
path := u.EscapedPath()
tokens := strings.Split(strings.Trim(path, "/"), "/")
filename = tokens[len(tokens)-1]
filePath := fmt.Sprintf("%s/%s_%s.js", jsPath, filename, uuid.NewV4().String())
``` 
Or if the objective is to ignore the file if the url is malformed do something like this:
```go
u, err := url.ParseRequestURI(URL)
	if err != nil {
            logger.Infof("skipping invalid url %s: %v" URL, err)
            return nil
	}
```

manelmontilla (changes requested):

cmd/vulcan-retirejs/main.go
julianvilas: Yes, you are totally right. When I started changing the code this part was in a function that wasn't returning errors, so I adapted the code to work for that situation. But now it doesn't make sense. Gonna apply your first suggestion.

julianvilas: Fixed in [eca3732](https://github.com/adevinta/vulcan-checks/pull/381/commits/eca3732bd6e9ba06b6e1847bf2a1efc20762a99e)

manelmontilla (approved):
LGTM

PR #386, opened 2023-07-25
Remove semgrep from arm blacklist

jesusfcr:
Semgrep released an arm64 image, so we can build our vulcan-semgrep arm64 check.


seilagamo (approved):
LGTM

PR #456, opened 2023-09-27
Fix and improve zap

jesusfcr:
- Improve speed/disk by disabling db recovery log (not a game changer)
  See <https://groups.google.com/g/zaproxy-users/c/L51nDFh0fjk>
- Ugrade github.com/zaproxy/zap-api-go
- Pin [2.13.0](https://www.zaproxy.org/blog/2023-07-12-zap-2.13.0/) and add to dependabot.
- Fix 2.13.0 not working because of add-on updates after server is suposed to be ready.
  ```
  level=warning msg="Scan not present in response body when calling Spider().Scan()" checkID=9f514c48-5d09-11ee-8109-465c9e98cb60"
  ```
- Add optimizations from [`zap_tune`](https://github.com/zaproxy/zaproxy/blob/5dd9a9f5c836d6c1e905265e3adc3e3c07d37467/docker/zap_common.py#L402)
- Setup max_scan_duration to allow the scan to report vulnerabilities before being killed by the agent.

### DIFF
--- a/cmd/vulcan-zap/main.go
+++ b/cmd/vulcan-zap/main.go
@@ -82,6 +82,7 @@ func main() {
 				"/zap/zap.sh",
 				"-daemon", "-host", "127.0.0.1", "-port", "8080",
 				"-config", "api.disablekey=true",
+				"-config", "database.recoverylog=false", // Reduce disk usage
 			).Output()
 			logger.Debugf("Error executing ZAP daemon: %v", err)
 			logger.Debugf("Output of the ZAP daemon: %s", string(out))
--- a/cmd/vulcan-zap/main.go
+++ b/cmd/vulcan-zap/main.go
@@ -81,7 +81,10 @@ func main() {
 				"-daemon", "-host", "127.0.0.1", "-port", "8080",
 				"-config", "api.disablekey=true",
 				"-config", "database.recoverylog=false", // Reduce disk usage
+				"-notel",  // Disables telemetry
+				"-silent", // Prevents from checking for addon updates
 			).Output()
+
 			logger.Debugf("Error executing ZAP daemon: %v", err)
 			logger.Debugf("Output of the ZAP daemon: %s", string(out))
 
--- a/cmd/vulcan-zap/main.go
+++ b/cmd/vulcan-zap/main.go
@@ -31,10 +31,7 @@ var (
 	client    zap.Interface
 )
 
-const (
-	contextName = "target"
-	contextID   = "1"
-)
+const contextName = "target"
 
 type options struct {
 	Depth    int     `json:"depth"`
@@ -56,7 +53,7 @@ type options struct {
 }
 
 func main() {
-	run := func(ctx context.Context, target, assetType, optJSON string, state checkstate.State) (err error) {
+	run := func(_ context.Context, target, assetType, optJSON string, state checkstate.State) (err error) {
 		var opt options
 		if optJSON != "" {
 			if err = json.Unmarshal([]byte(optJSON), &opt); err != nil {
@@ -127,9 +124,13 @@ func main() {
 			return fmt.Errorf("error parsing target URL: %w", err)
 		}
 
-		_, err = client.Context().NewContext(contextName)
+		cx, err := client.Context().NewContext(contextName)
 		if err != nil {
 			return fmt.Errorf("error creating scope context: %w", err)
+		}
+		contextID, err := getStringAttribute(cx, "contextId")
+		if err != nil {
+			return err
 		}
 
 		// Add base URL to the scope.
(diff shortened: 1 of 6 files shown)


cmd/vulcan-zap/main.go
```diff
@@ -213,11 +229,10 @@ func main() {
 				if err != nil {
 					return fmt.Errorf("error getting the status of the spider: %w", err)
 				}
-
 				v, ok := resp["status"]
 				if !ok {
 					// In this case if we can not get the status let's fail.
-					return errors.New("can not retrieve the status of the spider")
+					return fmt.Errorf("can not retrieve the status of the spider %v", resp)
```
jroimartin, line 235:
```suggestion
					return fmt.Errorf("cannot retrieve the status of the spider: %v", resp)
```

cmd/vulcan-zap/main.go
```diff
@@ -490,3 +505,15 @@ func isPluginIgnoredForFingerprint(opt options, pluginID string) bool {
 	}
 	return false
 }
+
+func getStringAttribute(m map[string]any, name string) (string, error) {
+	v, ok := m[name]
+	if !ok {
+		return "", fmt.Errorf("error %s not found", name)
```
jroimartin, line 512:
```suggestion
		return "", fmt.Errorf("%s not found", name)
```

cmd/vulcan-zap/main.go
```diff
@@ -490,3 +505,15 @@ func isPluginIgnoredForFingerprint(opt options, pluginID string) bool {
...
+	}
+	str, ok := v.(string)
+	if !ok {
+		return "", fmt.Errorf("error %s value [%v] is not a string", name, v)
```
jroimartin, line 516:
```suggestion
		return "", fmt.Errorf("%s value [%v] is not a string", name, v)
```

go.sum
jroimartin, line 1: I see two `zap-api-go` entries. Have you run `go mod tidy`?

cmd/vulcan-zap/manifest.toml
```diff
@@ -8,15 +8,18 @@ Timeout = 36000 # 10 hours. Expressed in seconds as an integer.
...
     "depth": 2, 
     "active": true, 
     "min_score": 0, 
     "disabled_scanners": ["10062", "10003", "10108"],
     "ignored_fingerprint_scanners": ["40018"],
     "max_spider_duration": 0,
-    "max_scan_duration": 0, 
+    "max_scan_duration": 540,
     "max_rule_duration": 0,
+    "max_alerts_per_rule": 10,
```
jroimartin, line 22: Does this mean that a rule will report at most 10 issues? If that's the case, we could potentially miss vulnerabilities, right?

jroimartin (changes requested):

jroimartin:
Overall LGTM. I just left some minor comments.

go.sum
jesusfcr: done, thanks

cmd/vulcan-zap/manifest.toml
jesusfcr: I understand this will cap the number of alerts a single rule can trigger. <https://github.com/zaproxy/zap-extensions/tree/main/addOns/pscanrules/src/main/java/org/zaproxy/zap/extension/pscanrules>, i.e. 
If a rule has been detected in 10 locations, it will stop evaluating that rule for the rest of the pages.
If that's the case I think this can be a good optimisation.

jroimartin: If it is at the risk of missing out vulnerabilities, I think this should be discussed further. Maybe I'm missing some context about Vulcan SaaS. @julianvilas @manelmontilla wdyt?

julianvilas: I don't know the context either in this case. @jesusfcr could you explain further the reason behind it?

jroimartin: For sure, we should change the name of the option to make clear that it only applies to the passive scan. Something like `pscan_max_alerts_per_rule`. `max_alerts_per_rule` is quite misleading and, actually, I had just reviewed the PR and forgot how it was used. Now taking into account that it only applies to passive scan rules, I understand that many alerts can be potentially the same, but not necessarily. For instance, [InformationDisclosureSuspiciousCommentsScanRule](https://github.com/zaproxy/zap-extensions/blob/main/addOns/pscanrules/src/main/java/org/zaproxy/zap/extension/pscanrules/InformationDisclosureSuspiciousCommentsScanRule.java) (I did not review all of them, this is just one rule I think could be detected multiple times without being the same problem).

manelmontilla: If we keep the option I agree with changing its name to something that mentions it applies only to passive scans, but in any case, from the context we have now I would not limit the number of alerts per rule by default, basically because the reason given by you @jroimartin; I prefer to not lose alerts than to have duplicates. Maybe what we can do is to leave the option but not enable it by default, as I can see this option to could be useful in the future in some repos using Lava.

jesusfcr: Reverted

jroimartin (approved):

PR #464, opened 2023-10-03
vulcan-trivy.go: improve Trivy CLI errors capture

julianvilas:
The Trivy CLI returns execution errors as log entries in STDERR even when the json output format is used. In fact the Trivy report data structure does not contain any field to store errors. That makes very hard to capture and process the Trivy CLI errors to be embedded in the 'check.Report.Error' attribute.

This commit tries to improve the situation by applying this changes:
- Reduce the output the Trivy CLI generates just to errors
- Ensure the error returned by the Trivy CLI is the one embedded in the 'check.Report.Error' attribute
- In case of Trivy CLI retries, make the check to log all of them but return just the last one
- Add an error tag to the check logs to make easier to filter logs by them. For example to filter logs by "retry trivy command execution failed"

### DIFF
--- a/cmd/vulcan-trivy/vulcan-trivy.go
+++ b/cmd/vulcan-trivy/vulcan-trivy.go
@@ -394,6 +394,9 @@ func execTrivy(opt options, action string, actionArgs []string) (*results, error
 		action,
 		"-f", "json",
 		"-o", reportOutputFile,
+		// Indicate Trivy CLI to output just errors to have a cleaner
+		// 'check.Report.Error'.
+		"--quiet",
 	}
 	// Show only vulnerabilities with specific severities.
 	if opt.Severities != "" {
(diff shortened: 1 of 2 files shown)


cmd/vulcan-trivy/vulcan-trivy.go
```diff
@@ -405,13 +408,16 @@ func execTrivy(opt options, action string, actionArgs []string) (*results, error
 
 	logger.Infof("running command: %s %s\n", trivyCmd, trivyArgs)
 
+	var inErr error
 	err := retry.Do(
```
jesusfcr, line 412: It's not from this PR but.
Do we need a retry? or a retry of 3 times? 
I think most of the checks doesn't have a retry mechanism in place.

jesusfcr:
Do you mind to remove the `time` from https://github.com/adevinta/vulcan-checks/blob/e8c8826cb82ff07c8fa21234bbd315fefe6d8a5d/cmd/vulcan-trivy/entrypoint.sh#L8
It's a leftover.

jesusfcr (approved):
LGTM

cmd/vulcan-trivy/vulcan-trivy.go
julianvilas: I thought the same when I saw it. Maybe @danfaizer can bright more light here.

danfaizer: As far as I remember, when I developed the check long ago, if the private registry was returning an error (for example 429 or 503) and wanted the check to retry.
However, now we know if the target is reachable or not (INCONCLUSIVE), therefore maybe we can remove the retry logic and in case we detect an increase of FAILED checks take action in the underlaying case of the error instead of just try to retry.

TL;DR
Perhaps yes: is better to remove the retry logic and log the Trivy command output accordingly in the error section of the report.

julianvilas: If you agree let's do one thing: I'll merge the PR as it is now as it will allow us to debug better the error causes. Then, I'll create a new PR removing the retry and if after the next round of scans we see an increase of errors related to network, we can just revert that commit. WDYT @jesusfcr @danfaizer?

jesusfcr: sure

jesusfcr (approved):

PR #469, opened 2023-10-06
Enable IP AssetType scanning for vulcan-nuclei check

danfaizer:
This PR enables IP AssetType scanning for vulcan-nuclei check.

For more detail visit internal discussion # 44.

### DIFF
--- a/cmd/vulcan-nuclei/local.toml.example
+++ b/cmd/vulcan-nuclei/local.toml.example
@@ -2,7 +2,7 @@
 
 [Check]
 Target = "https://example.com"
-AssetTypes = ["Hostname", "IP", "WebAddress"]
+AssetType = "WebAddress"
 # Available options example:
 # Options = """\
 #     {\
--- a/cmd/vulcan-nuclei/local.toml.example
+++ b/cmd/vulcan-nuclei/local.toml.example
@@ -2,7 +2,7 @@
 
 [Check]
 Target = "https://example.com"
-AssetType = "WebAddress"
+AssetTypes = ["Hostname", "IP", "WebAddress"]
 # Available options example:
 # Options = """\
 #     {\
--- a/cmd/vulcan-nuclei/manifest.toml
+++ b/cmd/vulcan-nuclei/manifest.toml
@@ -1,7 +1,7 @@
 # Copyright 2022 Adevinta
 
 Description = "Scan web addresses with projectdiscovery/nuclei"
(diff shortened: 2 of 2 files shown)


cmd/vulcan-nuclei/local.toml.example
```diff
@@ -2,7 +2,7 @@
 
 [Check]
 Target = "https://example.com"
-AssetType = "WebAddress"
+AssetTypes = ["Hostname", "IP", "WebAddress"]
```
jroimartin, line 5: Shouldn't a check define a single `AssetType`?

jroimartin (changes requested):

cmd/vulcan-nuclei/local.toml.example
danfaizer: Not really, a check may be capable of scanning different AssetTypes, [this](https://github.com/adevinta/vulcan-checks/blob/master/cmd/vulcan-tenable/manifest.toml#L3) is an example.

AssetTypes is used by the [vulcan-checks-bsys](https://github.com/adevinta/vulcan-checks-bsys/blob/master/manifest/manifest.go#L80-L82) in order to push to the `vulcan-persistence` the relation of which checks should run depending on the assettype.

So in fact, there was a typo in the `local.toml.example`.

jroimartin: Check the code I linked in my previous comment. If I'm not wrong a "check" is an instance of a "checktype" for a specific target, asset type and options. If a file `local.toml` exists, it is decoded into the check config and then the config is overriden by the provided [env vars](https://github.com/adevinta/vulcan-check-sdk/blob/f9b79af917a1cbc6feb3b86fc1cc0bc224229bab/config/config.go#L22-L27). This is useful when used in combination with the -t flag (test mode) to test the Vulcan check.

jroimartin: I've just realized I forgot to send the links:

- https://github.com/adevinta/vulcan-check-sdk/blob/f9b79af917a1cbc6feb3b86fc1cc0bc224229bab/config/config.go#L175-L185
- https://github.com/adevinta/vulcan-check-sdk/blob/f9b79af917a1cbc6feb3b86fc1cc0bc224229bab/config/config.go#L45-L52

jroimartin (approved):
LGTM

PR #549, opened 2024-02-05
Remove '--chown' from COPY instruction, as user no longer exists

kozmic:
The user 'gitleaks' was removed in this commit upstream: https://github.com/gitleaks/gitleaks/commit/ac4b5146b0f112df989b4374abb2b12799e37cba

When building this check now, it will fail with the error:
```
unable to convert uid/gid chown string to host mapping
```
Removing this argument pointing to a user that does not exists should fix the issue.

### DIFF
--- a/cmd/vulcan-gitleaks/Dockerfile
+++ b/cmd/vulcan-gitleaks/Dockerfile
@@ -11,6 +11,6 @@ ENTRYPOINT ["/usr/bin/env"]
 
 # Install check
 ARG TARGETOS TARGETARCH
-COPY --chown=gitleaks ${TARGETOS}/${TARGETARCH}/vulcan-gitleaks /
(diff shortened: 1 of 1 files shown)


jesusfcr (changes requested):
Hi, thanks for the PR

Do you mind to pin the base image to the latest `zricethezav/gitleaks:v8.18.2` in order to prevent this to happen again?

Later we should add the docker image to the dependabot config file and add the proper tests to allow safe upgrades.

kozmic:
@jesusfcr : Good suggestion, should be fixed in [86c6f49](https://github.com/adevinta/vulcan-checks/pull/549/commits/86c6f498c77cf1dcaff908f7b84d3f30834b31e9).

jesusfcr (approved):
LGTM, thanks!!

PR #550, opened 2024-02-07
Use aws-sdk v2 in the trusted advisor check

manelmontilla:
This PR introduces the following modifications to the vulcan-aws-trusted-advisor check: 
- Migrate the code to use the go AWS sdk v2. The v2 comes out of the box with a retrier that should prevent throttling errors.
- Avoid an out of bound access to a slice in some situations.
- Fixed some minor styling issues.
- Allow the check to get the AWS credentials from the process env vars, so it's easier to run it locally.


cmd/vulcan-aws-trusted-advisor/main.go
```diff
@@ -131,15 +121,35 @@ func scanAccount(opt options, target, assetType string, logger *logrus.Entry, st
 	if err != nil {
 		return err
 	}
-	creds, err := getCredentials(assumeRoleEndpoint, parsedARN.AccountID, role, logger)
-	if err != nil {
-		return err
+	var cfg aws.Config
```
jesusfcr, line 124: I think it would be better to move the reachability test after the get credentials part of the code.
Also validate that the account id you have access to is the target one.

jesusfcr (approved):

danfaizer (approved):
👍

PR #621, opened 2024-08-01
cmd/vulcan-govulncheck: add check

jroimartin:
This PR adds a new checktype that uses govulncheck to report known
vulnerabilities that affect Go code.


cmd/vulcan-govulncheck/main.go
```diff
@@ -0,0 +1,208 @@
...
+			}
+		}
+		rscs = append(rscs, rsc)
+	}
+
+	location := getLocations(result)
+	vuln := report.Vulnerability{
+		Summary:          rule.ShortDescription.Text,
+		Score:            calcScore(result),
+		AffectedResource: location,
```
danfaizer, line 178: AffectedResource refers to the specific and unique resource impacted by the finding. If there are multiple locations, we may create one vulnerability for each location (e.g., repositories with submodules). Testing this would be convenient, though I'm not sure if it's possible.

If the AffectedResource format isn't user-friendly, you can use the AffectedResourceString attribute to present the information to the user.

More details [here](https://github.com/adevinta/vulcan-report/blob/master/report.go#L117-L118).

cmd/vulcan-govulncheck/main.go
```diff
@@ -0,0 +1,208 @@
...
+		Fingerprint:      helpers.ComputeFingerprint(result.RuleID, location),
```
danfaizer, line 179: The `location` is used to compute the fingerprint, which is fine. However, if there can be more than one location and the order is not deterministic, it would be beneficial to sort the slice to ensure the fingerprint is deterministic.

cmd/vulcan-govulncheck/main.go
```diff
@@ -0,0 +1,208 @@
...
+	var locs []string
+	for _, loc := range result.Locations {
+		locs = append(locs, loc.PhysicalLocation.String())
+	}
+	return strings.Join(locs, " ")
+}
+
+func calcScore(result sarif.Result) float32 {
+	if result.Level == "error" {
+		return 8 // high severity
```
danfaizer, line 204: As `vulcan-report` package is already imported, I may suggest to use [report.SeverityThresholdHigh](https://github.com/adevinta/vulcan-report/blob/master/utils.go#L37) in order to return a High vulnerability score.

cmd/vulcan-govulncheck/main.go
```diff
@@ -0,0 +1,208 @@
...
+	} else {
+		return 0 // informational severity
```
danfaizer, line 206: As `vulcan-report` package is already imported, I may suggest to use [report.SeverityThresholdNone](https://github.com/adevinta/vulcan-report/blob/master/utils.go#L31) in order to return a Info vulnerability score.

danfaizer (changes requested):
Thank you for the PR.
I've added some comments. Please feel free to reach out if you need further context.

cmd/vulcan-govulncheck/main.go
jroimartin: I like it. It makes the code self-explanatory.

cmd/vulcan-govulncheck/main.go
jroimartin: I like it. It makes the code self-explanatory. Fixed.

cmd/vulcan-govulncheck/main.go
jroimartin: I agree. We should be defensive here and do not rely on govulncheck's output to be always one location or multiple locations with the same order. Fixed.

cmd/vulcan-govulncheck/main.go
danfaizer: The refactor done looks better and more accurate.
Thanks!

danfaizer (approved):
👍

PR #719, opened 2024-10-28
Update trivy-check Dockerfile [ change container registry source for images ] 

muhammedabdelkader:
change repos to use AWS ecr aqauasecurity, because ghcr needs authented to download db and java-db containers


kozmic:
This solution is discussed here: https://github.com/aquasecurity/trivy/discussions/7668#discussioncomment-11028887

jesusfcr (approved):

PR #790, opened 2025-01-22
Vulcan subdomain takeover

seilagamo:


cmd/vulcan-subdomain-takeover/main.go
```diff
@@ -0,0 +1,381 @@
...
+				}
+				return nil, fmt.Errorf("getting AWS metadata for %s: %w", record, err)
+			}
+			for _, service := range strings.Split(awsMetadata.Service, ",") {
+				if service == "EC2" {
+					dnsEC2IPs[dnsr.name] = dnsr.records
+				}
+			}
+		}
+	}
```
jesusfcr, line 313: Why not just use a regex.MustCompile(`\bEC2\b`) for testing that without a loop?

cmd/vulcan-subdomain-takeover/main.go
```diff
@@ -0,0 +1,381 @@
...
+		logger:         logger,
+		target:         target,
+		route53Client:  route53.NewFromConfig(cfg),
+		ec2Client:      ec2.NewFromConfig(cfg),
+		ipRangesClient: NewAWSIPRanges(),
+	}, nil
+}
+
+// Run executes the scan.
+func (s Scanner) Run() (map[string]string, error) {
```
jesusfcr, line 109: The process has an issue because the DNS records are global, but the getIPs is just for one region.
So it would report a False positive if one DNS record points to an IP on other region.

jesusfcr (changes requested):
Great job.

I added some minor comments, but the main concern is that it doesn't take into consideration multi region.

Apart from that I think the name of the check is too generic for the current implementation. One posibility could be to change to `vulcan-aws-subdomain-takeover`. Other posibility could be to include this functionality in the current `vulcan-aws-alerts` check.

Apart from that, I probably base the PR on master and not to checkshttp to prevent blockings.

cmd/vulcan-govulncheck/Dockerfile
```diff
@@ -1,7 +1,7 @@
 # Copyright 2024 Adevinta
 
-FROM golang:1.22.6-alpine3.20
-ENV GOVULNCHECK_VERSION=v1.1.3
```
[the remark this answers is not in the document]
seilagamo: It seems not related, but since I am using 1.23 features, the old version of govulncheck started to fail in the release github action and I had to update the version. You can see an example of this [here](https://github.com/adevinta/vulcan-checks/actions/runs/13028991120/job/36343868907)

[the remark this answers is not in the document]
jesusfcr: Ok then.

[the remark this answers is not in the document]
seilagamo: Finally, I extracted this update in other PR: https://github.com/adevinta/vulcan-checks/pull/799

cmd/vulcan-aws-subdomain-takeover/cloudinventory.go
```diff
@@ -0,0 +1,72 @@
...
+		token:    token,
+		endpoint: endpoint,
+	}
+}
+
+func (ci *CloudInventory) IsIPPublicInInventory(ip string) (bool, error) {
+	t, err := template.New("isInInventory").Parse(ci.endpoint)
+	if err != nil {
+		return false, err
+	}
```
jesusfcr, line 33: Move this logic to `NewCloudInventory` and store the parsed template in `CloudInventory`

jesusfcr (dismissed):
Also  have a look to the issues from `golangci-lint run cmd/vulcan-aws-subdomain-takeover`

cmd/vulcan-aws-subdomain-takeover/awsipranges.go
```diff
@@ -0,0 +1,200 @@
...
+
+// ErrPrefixNotFound means tha a prefix wasn't found in the ip ranges.
+var ErrPrefixNotFound = errors.New("prefix not found")
+
+// AWSIPRanges contains two strings, SyncToken and CreateDate, and a slice of Prefixes.
+type AWSIPRanges struct {
+	SyncToken  string `json:"syncToken"`
+	CreateDate string `json:"createDate"`
+	d          ipRangesRetriever
+	IPPrefixes []IPPrefix `json:"prefixes" json_alias:"ip6_prefixes"`
```
jesusfcr, line 35: What is this json_alias:"ip6_prefixes" ?

seilagamo: If you look into the file https://ip-ranges.amazonaws.com/ip-ranges.json, you can see there are two groups of ips, IP4 and IP6. The JSON structure is basically the same, only the key changes. Using the alias you can unmarshal those two JSON structures into the same golang struct type.

cmd/vulcan-aws-subdomain-takeover/local.toml.example
```diff
@@ -0,0 +1,10 @@
+[Check]
+Target = "arn:aws:iam::123456789012:root"
+AssetType = "AWSAccount"
+Options = '{"global": true}'
+
+[RequiredVars]
+VULCAN_ASSUME_ROLE_ENDPOINT="http://localhost:8080/assume"
+ROLE_NAME="SecurityAuditRole"
+CLOUD_INVENTORY_TOKEN=
+CLOUD_INVENTORY_ENDPOINT="http://localhost:8081/api/publicips/{{.IP}}
```
jesusfcr, line 10: To make this less custom and provide more flexibility I would make those changes:

```suggestion
INVENTORY_ENDPOINT="http://localhost:8081/api/publicips/{{.IP}}"      # Optional and I would remove the Global option.
INVENTORY_HEADERS='["token: MYSECRETTOKEN", "Content-Type: application/json"]'    # Optional headers
INVENTORY_NOTFOUND_BODY='{}'    # Optional regex for not found. If empty just rely on the status.
```

jesusfcr (changes requested):

jesusfcr (changes requested):
I added some comments.

cmd/vulcan-aws-subdomain-takeover/main.go
```diff
@@ -0,0 +1,391 @@
...
+}
+
+// ipRangesClient represents a IPRanges Client.
+type ipRangesClient interface {
+	GetPrefixes() (AWSPrefixes, error)
+}
+
+// Scanner represents a subdomain takeover scanner.
+type Scanner struct {
+	inventory      Inventory
```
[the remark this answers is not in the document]
seilagamo: Inventory is an interface; you don't have to use a pointer there

jesusfcr (approved):
LGTM. Great job!

PR #799, opened 2025-02-03
Update govulncheck version

seilagamo:
Update the version of govulncheck to the last one.


jesusfcr (approved):
LGTM

I would add the check for automatic upgrades, but it's up to you:

- Add the check to .github/dependabot.yaml.
- Add a test.sh to the check. (See the semgrep one). I know we should migrate to lava but not in this pr.

Also, rebase the PR.

jesusfcr (approved):
LGTM

PR #888, opened 2025-06-20
ghe-reposec: change dependabot check

danfaizer:
Change the endpoint to verify if Dependabot (automatic security fixes) is enabled and remove GHAS controls.


jesusfcr (approved):
LGTM

